Issue #24, opened 2017-09-14
Cannot set the breakpoint for file from included path when restart debug session


yaohaizh:
Project: https://github.com/fbricon/vscode-java-demo
System: Ubuntu

1. Git clone the project
2. Open with VSCode and wait for the project initialized with language server
3. Config launch.json
4. Start debugging

Issue:  
1. Debug into SpringApplication which is not part of the workspace but in the included path
2. Set a breakpoint on the SprintApplication
3. Restart debug session. 

Cannot set the breakpoint for SpringApplication

```console
[Error - 11:22:49 PM] Sep 13, 2017 11:21:52 PM DebugSession dispatch exception:java.lang.IllegalStateException
[Error - 11:37:19 PM] Sep 13, 2017 11:21:52 PM File not found: '/vscode-java-demo/SpringApplication.java'
Resource '/vscode-java-demo/SpringApplication.java' does not exist.
org.eclipse.core.internal.resources.ResourceException: Resource '/vscode-java-demo/SpringApplication.java' does not exist.
	at org.eclipse.core.internal.resources.Resource.checkExists(Resource.java:335)
	at org.eclipse.core.internal.resources.Resource.checkAccessible(Resource.java:209)
	at org.eclipse.core.internal.resources.File.getContents(File.java:275)
	at org.eclipse.jdt.internal.core.util.Util.getResourceContentsAsCharArray(Util.java:1190)
	at org.eclipse.jdt.internal.core.CompilationUnit.getContents(CompilationUnit.java:652)
	at org.eclipse.jdt.core.dom.ASTParser.internalCreateAST(ASTParser.java:1125)
	at org.eclipse.jdt.core.dom.ASTParser.createAST(ASTParser.java:813)
	at com.microsoft.java.debug.plugin.internal.JdtSourceLookUpProvider.getFullyQualifiedName(JdtSourceLookUpProvider.java:103)
	at com.microsoft.java.debug.core.adapter.handler.SetBreakpointsRequestHandler.convertClientBreakpointsToDebugger(SetBreakpointsRequestHandler.java:110)
	at com.microsoft.java.debug.core.adapter.handler.SetBreakpointsRequestHandler.handle(SetBreakpointsRequestHandler.java:76)
	at com.microsoft.java.debug.core.adapter.DebugAdapter.dispatchRequest(DebugAdapter.java:72)
	at com.microsoft.java.debug.core.adapter.ProtocolServer.dispatchRequest(ProtocolServer.java:201)
[...]

PR #58, opened 2017-09-29
Support uri in Source.path

testforstephen:

This feature will also resolve these bugs: 
https://github.com/Microsoft/java-debug/issues/24
https://github.com/Microsoft/vscode-java-debug/issues/46
https://github.com/Microsoft/vscode-java-debug/issues/21


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/AdapterUtils.java
```diff
@@ -151,6 +153,19 @@ public static String toUri(String path) {
     }
 
     /**
+     * Check a string variable is an uri or not.
```
yaohaizh: an ==> a

PR #68, opened 2017-10-09
add log for catch (Exception) clause 

andxu:
…exception there.


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Log.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public static void setLevel(Level newLevel) {
+        logger.setLevel(newLevel);
+    }
+
+    /**
+     * Add a log Handler to receive logging messages.
+     * @param   handler a logging Handler
+     */
+    public static void addHandler(Handler handler) {
```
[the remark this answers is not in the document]
andxu: added

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Log.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * This helper class provides static methods for a default instance of <code>java.util.logging.Logger</code>,a simple code like
+ * <code>Log.info("Sample trace: %d", 10)</code> is able to replace <code>logger.info(String.format("Sample trace: %d", 10))</code>.
+ * Using a very short name to better code fitness.
+ */
+public final class Log {
```
akaroml: Why do we need to wrap around the JDK logger? I didn't see much value added. Am I missing anything?

andxu: This PR is mainly to focus on simplifying the logging code to better code.

andxu: Trace user data is also include in Log.java, Log will be the only class need to record logs and user datas.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/ProtocolServer.java
```diff
@@ -86,30 +83,29 @@ public ProtocolServer(InputStream input, OutputStream output, IProviderContext c
...
                     break;
                 }
 
-                this.rawData.append(new String(buffer, 0, read).getBytes(PROTOCOL_ENCODING));
+                rawData.append(new String(buffer, 0, read).getBytes(PROTOCOL_ENCODING));
                 this.processData();
             }
         } catch (IOException e) {
-            logger.log(Level.SEVERE, String.format("Read data from io exception: %s", e.toString()), e);
+            Log.error("Read data from io exception: %s", e.toString());
```
testforstephen: should Log.endSession at the end of the start method.

Issue #41, opened 2017-09-21
lambda expression doesn't apply to the right java file.


andxu:
Step: use debugger to debug itself, step in computeIfAbsent in StackTraceRequestHandler.java@112

![image](https://user-images.githubusercontent.com/17075863/30673467-e8f0682c-9ea5-11e7-9326-287ae42fb4fc.png)

testforstephen:
The lambda expression issue has been resolved. As to the improvement of method names in call stack view is duplicated with https://github.com/Microsoft/java-debug/issues/43

PR #72, opened 2017-10-12
Improve the call stack display info for not avialable source

testforstephen:


testforstephen:
fix the bug https://github.com/Microsoft/java-debug/issues/41

Issue #43, opened 2017-09-21
Include class name in stack frame


andxu:
Currently only the method is displayed:

![image](https://user-images.githubusercontent.com/17075863/30675009-8cdcd880-9eb0-11e7-9a41-7c004f6a209e.png)

see eclipse:
![image](https://user-images.githubusercontent.com/17075863/30675017-990c78b8-9eb0-11e7-9f96-b14c329a9a70.png)

testforstephen:
The code has merged to master branch.

PR #73, opened 2017-10-12
format stackframe name with ClassName.MethodName(ParameterTypes)

testforstephen:

add for feature https://github.com/Microsoft/java-debug/issues/43

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StackTraceRequestHandler.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StackTraceRequestHandler.java
@@ -15,6 +15,7 @@ import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


testforstephen:
image

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StackTraceRequestHandler.java
```diff
@@ -152,4 +154,21 @@ public void handle(Command command, Arguments arguments, Response response, IDeb
             }
         }
     }
+
+    private String formatMethodName(Method method, boolean showContextClass, boolean showParameter) {
+        String formattedName = method.name();
```
yaohaizh: Better to use StringBuilder in this method for lots of String concat

andxu: why not use typeformatter

testforstephen: fixed.

PR #84, opened 2017-10-18
fix the inconsistent uri issue for set breakpoint request

testforstephen:

Handle the vscode bug: https://github.com/Microsoft/vscode/issues/36471

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/AdapterUtils.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/AdapterUtils.java
@@ -11,8 +11,10 @@
 
 package com.microsoft.java.debug.core.adapter;
 
+import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystemNotFoundException;
 import java.nio.file.Files;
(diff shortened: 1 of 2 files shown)


yaohaizh:
Will this target 0.2.0 release?

testforstephen:
i prefer to include it in 0.2.0

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/AdapterUtils.java
```diff
@@ -220,4 +222,18 @@ public static String getSHA256HexDigest(String content) {
...
+
+    /**
+     * decode the uri string.
+     * @param uri
+     *          the uri string
+     * @return the decoded uri
+     */
+    public static String decodeURIComponent(String uri) {
+        try {
+            return URLDecoder.decode(uri, "UTF-8");
```
yaohaizh: Please use the constant String definition if can.  
```
public static String ... 
```

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/AdapterUtils.java
```diff
@@ -220,4 +222,18 @@ public static String getSHA256HexDigest(String content) {
         }
         return buf.toString();
     }
+
+    /**
+     * decode the uri string.
```
yaohaizh: `decode` ==> `Decode`

Issue #49, opened 2017-09-25
Perf issue on get locations of breakpoint


andxu:
![image](https://user-images.githubusercontent.com/17075863/30804144-e53ef6e0-a21e-11e7-8246-d5b6a27608a8.png)

PR #88, opened 2017-10-24
Fix perf issue on get locations of breakpoint

testforstephen:
fix bug #49.

Nested types are just needed for those already loaded classes. For the referenceType from ClassPrepareEvent, no need to search the locations in the nested types. For newly loaded classes, the debugger has also registered ClassPrepareRequest for their local types, so the debugger will handle the local types in ClassPrepareEvent handler.

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
@@ -133,8 +133,8 @@ public class Breakpoint implements IBreakpoint {
                             || localClassPrepareRequest.equals(debugEvent.event.request())))
                 .subscribe(debugEvent -> {
                     ClassPrepareEvent event = (ClassPrepareEvent) debugEvent.event;
-                    List<BreakpointRequest> newRequests = createBreakpointRequests(event.referenceType(),
-                            lineNumber, hitCount);
+                    List<BreakpointRequest> newRequests = createBreakpointRequests(event.referenceType(), lineNumber,
+                            hitCount, false);
                     requests.addAll(newRequests);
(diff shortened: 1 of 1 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
```diff
@@ -167,30 +168,44 @@ public void setHitCount(int hitCount) {
...
+    private static List<Location> collectLocations(List<ReferenceType> refTypes, int lineNumber, boolean includeNestedTypes) {
         List<Location> locations = new ArrayList<>();
         try {
             refTypes.forEach(refType -> {
-                locations.addAll(collectLocations(refType, lineNumber));
-                locations.addAll(collectLocations(refType.nestedTypes(), lineNumber));
+                List<Location> newLocations = collectLocations(refType, lineNumber);
+                if (!newLocations.isEmpty()) {
+                    locations.addAll(newLocations);
+                } else if (includeNestedTypes) {
```
ansyral: just if?

testforstephen: what's the advice?

ansyral: i mean why here it is `else if` rather than `if`?

testforstephen: if the breakpoint located in the refType, no need to search the nested type.

Goom11:
Hello, @testforstephen, how do you know that ReferenceType.nestedTypes invokes VirtualMachine.allClasses?

PR #94, opened 2017-11-06
add user settings

andxu:


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaDebugDelegateCommandHandler.java
```diff
@@ -32,6 +33,8 @@
 
     public static String CONFIG_LOG_LEVEL = "vscode.java.configLogLevel";
 
+    public static String UPDATE_USER_SETTINGS = "vscode.java.updateUserSettings";
```
akaroml: We need to not confuse the user by providing the command vscode.java.updateUserSetting. It's actually debugger settings. So suggest making it "vscode.java.updateDebuggerSettings".

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/formatter/StringObjectFormatter.java
```diff
@@ -26,7 +26,7 @@
 
 public class StringObjectFormatter extends ObjectFormatter implements IValueFormatter {
     public static final String MAX_STRING_LENGTH_OPTION = "max_string_length";
-    private static final int DEFAULT_MAX_STRING_LENGTH = 100;
+    private static final int DEFAULT_MAX_STRING_LENGTH = 200;
```
akaroml: Can I understand why 100 -> 200?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/EvaluateRequestHandler.java
```diff
@@ -61,18 +59,11 @@ public void handle(Command command, Arguments arguments, Response response, IDeb
...
+        final boolean showStaticVariables = UserSettings.showStaticVariables;
+
         Map<String, Object> options = context.getVariableFormatter().getDefaultOptions();
-        if (evalArguments.format != null && evalArguments.format.hex) {
-            options.put(NumericFormatter.NUMERIC_FORMAT_OPTION, NumericFormatEnum.HEX);
-        }
-        if (showFullyQualifiedNames) {
-            options.put(SimpleTypeFormatter.QUALIFIED_CLASS_NAME_OPTION, showFullyQualifiedNames);
-        }
+        VariableUtils.applyFormatterOptions(options, evalArguments.format != null && evalArguments.format.hex,
```
akaroml: Do we need to apply the formatter options every time a evaluate request is processed? If the options stay inside VadiableUtils, then we can call it when a user actually changes the settings.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/VariablesRequestHandler.java
```diff
@@ -57,18 +55,11 @@ public void handle(Command command, Arguments arguments, Response response, IDeb
...
-        boolean showStaticVariables = true;
-        // TODO: When vscode protocol support customize settings of value format, showFullyQualifiedNames should be one of the options.
-        boolean showFullyQualifiedNames = true;
-        if (varArgs.format != null && varArgs.format.hex) {
-            options.put(NumericFormatter.NUMERIC_FORMAT_OPTION, NumericFormatEnum.HEX);
-        }
-        if (showFullyQualifiedNames) {
-            options.put(SimpleTypeFormatter.QUALIFIED_CLASS_NAME_OPTION, showFullyQualifiedNames);
-        }
+        VariableUtils.applyFormatterOptions(options, varArgs.format != null && varArgs.format.hex, false);
```
akaroml: Same here, do we really need to apply settings during processing a request?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/formatter/StringObjectFormatter.java
andxu: the default length is the length most users saw, the 100 is too small for normal cases like connections string, path, I would advise to use 256.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/EvaluateRequestHandler.java
andxu: The condition for generating options is also influenced by argument, not only user settings.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/DebugSettings.java
```diff
@@ -0,0 +1,102 @@
...
+                    String jsonStr = (String) arg;
+
+                    try {
+                        Map<String, Object> map = JsonUtils.fromJson(jsonStr, new HashMap<String, Object>().getClass());
+
+                        for (String keyStr : map.keySet()) {
+                            Object valueObj = map.get(keyStr);
+                            try {
+                                switch (keyStr) {
+                                    case "show_hex":
```
testforstephen: Not consistent with the key on vscode user settings.

Issue #98, opened 2017-11-10
java.util.ConcurrentModificationException in toGson


andxu:
!ENTRY java-debug 4 0 2017-11-10 09:31:50.091
!MESSAGE Dispatch debug protocol error: java.util.ConcurrentModificationException
!STACK 0
java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.write(CollectionTypeAdapterFactory.java:96)
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.write(CollectionTypeAdapterFactory.java:61)
	at com.google.gson.Gson.toJson(Gson.java:669)
	at com.google.gson.Gson.toJson(Gson.java:648)
	at com.google.gson.Gson.toJson(Gson.java:603)
	at com.google.gson.Gson.toJson(Gson.java:583)
	at com.microsoft.java.debug.core.protocol.JsonUtils.toJson(JsonUtils.java:42)
	at com.microsoft.java.debug.core.UsageDataSession.submitUsageData(UsageDataSession.java:140)
	at com.microsoft.java.debug.core.adapter.ProtocolServer.stop(ProtocolServer.java:62)
	at com.microsoft.java.debug.core.adapter.ProtocolServer.dispatchRequest(ProtocolServer.java:68)
	at com.microsoft.java.debug.core.protocol.AbstractProtocolServer.processData(AbstractProtocolServer.java:166)
	at com.microsoft.java.debug.core.protocol.AbstractProtocolServer.start(AbstractProtocolServer.java:80)
	at com.microsoft.java.debug.core.adapter.ProtocolServer.start(ProtocolServer.java:53)
	at com.microsoft.java.debug.plugin.internal.JavaDebugServer$2.run(JavaDebugServer.java:136)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)

andxu:
The code is checked in

PR #103, opened 2017-11-14
add synchronize statement for modifying eventlist and serializing it

andxu:
…ntlist.


andxu:
https://github.com/Microsoft/java-debug/issues/98

PR #112, opened 2017-11-20
[Refactor] use CompletableFuture to support async response in debug adapter

testforstephen:

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
@@ -68,7 +68,7 @@ public class DebugAdapter implements IDebugAdapter {
         try {
             if (debugContext.isVmTerminated()) {
                 // the operation is meaningless
-                return AdapterUtils.createAsyncResponse(response);
+                return CompletableFuture.completedFuture(response);
             }
             List<IDebugRequestHandler> handlers = requestHandlers.get(command);
             if (handlers != null && !handlers.isEmpty()) {
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
@@ -78,9 +78,6 @@ public class DebugAdapter implements IDebugAdapter {
                         return result;
                     }
                 }
-            } else {
-                return AdapterUtils.createAsyncErrorResponse(response, ErrorCode.UNRECOGNIZED_REQUEST_FAILURE,
-                        String.format("Unrecognized request: { _request: %s }", request.command));
             }
         } catch (Exception e) {
             logger.log(Level.SEVERE, String.format("DebugSession dispatch exception: %s", e.toString()), e);
@@ -88,7 +85,8 @@ public class DebugAdapter implements IDebugAdapter {
                     e.getMessage() != null ? e.getMessage() : e.toString());
         }
 
-        return AdapterUtils.createAsyncResponse(response);
+        return AdapterUtils.createAsyncErrorResponse(response, ErrorCode.UNRECOGNIZED_REQUEST_FAILURE,
+                String.format("Unrecognized request: { _request: %s }", request.command));
     }
 
     private void initialize() {
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
@@ -65,28 +65,24 @@ public class DebugAdapter implements IDebugAdapter {
         Command command = Command.parse(request.command);
         Arguments cmdArgs = JsonUtils.fromJson(request.arguments, command.getArgumentType());
 
-        try {
-            if (debugContext.isVmTerminated()) {
-                // the operation is meaningless
-                return CompletableFuture.completedFuture(response);
-            }
-            List<IDebugRequestHandler> handlers = requestHandlers.get(command);
-            if (handlers != null && !handlers.isEmpty()) {
-                for (IDebugRequestHandler handler : handlers) {
-                    CompletableFuture<Messages.Response> result = handler.handle(command, cmdArgs, response, debugContext);
-                    if (result != null) {
-                        return result;
-                    }
-                }
-            }
-        } catch (Exception e) {
-            logger.log(Level.SEVERE, String.format("DebugSession dispatch exception: %s", e.toString()), e);
-            return AdapterUtils.createAsyncErrorResponse(response, ErrorCode.UNKNOWN_FAILURE,
-                    e.getMessage() != null ? e.getMessage() : e.toString());
+        if (debugContext.isVmTerminated()) {
+            // the operation is meaningless
+            return CompletableFuture.completedFuture(response);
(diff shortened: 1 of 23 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/ThreadsRequestHandler.java
```diff
@@ -45,36 +46,29 @@
...
-                this.pause((PauseArguments) arguments, response, context);
-                break;
+                return this.pause((PauseArguments) arguments, response, context);
             case CONTINUE:
-                this.resume((ContinueArguments) arguments, response, context);
-                break;
+                return this.resume((ContinueArguments) arguments, response, context);
             default:
-                return;
+                return null;
```
akaroml: Instead of returning null, try to complete the future with an exception by calling future.completeExceptionally

akaroml: Or return AdapterUtils.createAsyncErrorResponse

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/ProtocolServer.java
```diff
@@ -51,24 +45,28 @@ public ProtocolServer(InputStream input, OutputStream output, IProviderContext c
...
-        super.stop();
         usageDataSession.submitUsageData();
     }
 
+    @Override
+    protected void sendMessage(Messages.ProtocolMessage message) {
+        super.sendMessage(message);
+        if (message instanceof Messages.Response) {
+            usageDataSession.recordResponse((Messages.Response) message);
+        } else if (message instanceof Messages.Request) {
```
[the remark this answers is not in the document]
testforstephen: YES, when implementing launchInTerminal, the debugger will send a request to VSCode.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/ProtocolServer.java
```diff
@@ -51,24 +45,28 @@ public ProtocolServer(InputStream input, OutputStream output, IProviderContext c
     public void start() {
         usageDataSession.reportStart();
         super.start();
-    }
-
-    /**
-     * Sets terminateSession flag to true. And the dispatcher loop will be terminated after current dispatching operation finishes.
-     */
-    public void stop() {
         usageDataSession.reportStop();
```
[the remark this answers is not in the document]
testforstephen: It has the same intention.
The start() is a while-loop, when the debug session is stopped, the start method will exit and report the usage data there.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/ProtocolServer.java
```diff
@@ -51,24 +45,28 @@ public ProtocolServer(InputStream input, OutputStream output, IProviderContext c
...
+        }
+    }
+
     protected void dispatchRequest(Messages.Request request) {
-        Messages.Response response = this.debugAdapter.dispatchRequest(request);
-        if (request.command.equals("disconnect")) {
-            stop();
+        usageDataSession.recordRequest(request);
+        CompletableFuture<Messages.Response> future = this.debugAdapter.dispatchRequest(request);
+        if (future != null) {
```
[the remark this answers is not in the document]
testforstephen: Here i found no need to check null because the debug adapter will always send a future back.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/ThreadsRequestHandler.java
testforstephen: nice.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/ProtocolServer.java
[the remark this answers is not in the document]
akaroml: Got it, thanks.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/ProtocolServer.java
```diff
@@ -33,42 +33,36 @@
...
         }
-        sendMessage(response);
-        usageDataSession.recordResponse(response);
+    }
+
+    protected void dispatchRequest(Messages.Request request) {
+        usageDataSession.recordRequest(request);
+        this.debugAdapter.dispatchRequest(request).thenAccept((response) -> {
+            sendMessage(response);
+        });
```
akaroml: We need to have some code to handle exception from the promise. A promise isn't always completed. It can fail with an exception too.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
```diff
@@ -72,46 +68,25 @@ public DebugAdapter(BiConsumer<Events.DebugEvent, Boolean> consumer, IProviderCo
...
             if (debugContext.isVmTerminated()) {
                 // the operation is meaningless
-                return response;
+                return CompletableFuture.completedFuture(response);
             }
             List<IDebugRequestHandler> handlers = requestHandlers.get(command);
             if (handlers != null && !handlers.isEmpty()) {
                 for (IDebugRequestHandler handler : handlers) {
-                    handler.handle(command, cmdArgs, response, debugContext);
+                    CompletableFuture<Messages.Response> result = handler.handle(command, cmdArgs, response, debugContext);
```
akaroml: Suggest also guard the handle call in future like this:
return CompletableFuture.supplyAsync(() -> future.get())

So whatever happens in the handler won't affect the request loop. For example, we don't even need to check whether the result is null because it will automatically throw a NPE.

testforstephen: Fixed.
`CompletableFuture.thenCompose` is more suitable for our scenario.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/ProtocolServer.java
testforstephen: fixed.
Use `future.whenComplete()` to handle normal response and exceptional response.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
```diff
@@ -65,28 +65,24 @@ public DebugAdapter(Consumer<Messages.ProtocolMessage> sendMessage, IProviderCon
...
-                }
+        if (debugContext.isVmTerminated()) {
+            // the operation is meaningless
+            return CompletableFuture.completedFuture(response);
+        }
+        List<IDebugRequestHandler> handlers = requestHandlers.get(command);
+        if (handlers != null && !handlers.isEmpty()) {
+            CompletableFuture<Messages.Response> future = CompletableFuture.completedFuture(response);
+            for (IDebugRequestHandler handler : handlers) {
+                future = future.thenCompose((res) -> {
```
akaroml: This is PERFECT! thanks!

PR #119, opened 2017-11-28
Auto add correlationId in usage data

Eskibear:
- Add `AmbientContext` class, using InheritableThreadLocal to store context data including correlationId

- CorrelationId format: `[GUID of VSCode client/window].[Sequence # of debug session]`, e.g. `825b242b-7a73-4f5f-bcda-ab6eee88e28d.6` means the user's 6th launch of debug session in current window.

By doing this, we now know relation of these debug sessions.

image

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/UsageDataSession.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/UsageDataSession.java
@@ -20,18 +20,18 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import com.google.gson.JsonElement;
+import com.microsoft.AmbientContext;
 import com.microsoft.java.debug.core.adapter.AdapterUtils;
 import com.microsoft.java.debug.core.protocol.JsonUtils;
 import com.microsoft.java.debug.core.protocol.Messages.Request;
 import com.microsoft.java.debug.core.protocol.Messages.Response;
-import com.microsoft.java.debug.core.trace.AmbientContext;
 import com.sun.jdi.event.Event;
 
 public class UsageDataSession {
     private static final Logger logger = Logger.getLogger(Configuration.LOGGER_NAME);
     private static final Logger usageDataLogger = Logger.getLogger(Configuration.USAGE_DATA_LOGGER_NAME);
     private static final long RESPONSE_MAX_DELAY_MS = 1000;
-    private static final String AMBCTX_USAGE_DATA_SESSION = "UsageDataSession";
+    private static final String AMBIENT_CONTEXT_USAGE_DATA_SESSION = "UsageDataSession";
 
     private final String sessionGuid = UUID.randomUUID().toString();
     private boolean jdiEventSequenceEnabled = false;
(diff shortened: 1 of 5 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/trace/AmbientContext.java
```diff
@@ -0,0 +1,90 @@
...
+
+package com.microsoft.java.debug.core.trace;
+
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class AmbientContext extends ConcurrentHashMap<String, Object> {
+    public static final String AMBCTX_ID = AmbientContext.class.getSimpleName() + "Id";
+    static InheritableThreadLocal<AmbientContext> callContext = new InheritableThreadLocal<AmbientContext>() {
```
akaroml: Do we really want to use the name callContext here? This is implying something which behaves just like the .net CallContext.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/trace/AmbientContext.java
```diff
@@ -0,0 +1,90 @@
...
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Eclipse Public License v1.0
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v10.html
+*
+* Contributors:
+*     Microsoft Corporation - initial API and implementation
+*******************************************************************************/
+
+package com.microsoft.java.debug.core.trace;
```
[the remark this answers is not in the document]
Eskibear: Considering the scenarios here in java-debug (context data like correlationId only flow from upstream to downstream), I've removed most functionality of the original AmbientContext for simplicity.  
Actually independent package like `com.microsoft.AmbientContext` was what I implemented at very beginning, which was almost a copy of `System.something`. Together with `AmbientContextExtension`, it just looks heavy.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/UsageDataSession.java
```diff
@@ -24,13 +24,14 @@
...
 import com.microsoft.java.debug.core.protocol.Messages.Response;
+import com.microsoft.java.debug.core.trace.AmbientContext;
 import com.sun.jdi.event.Event;
 
 public class UsageDataSession {
     private static final Logger logger = Logger.getLogger(Configuration.LOGGER_NAME);
     private static final Logger usageDataLogger = Logger.getLogger(Configuration.USAGE_DATA_LOGGER_NAME);
     private static final long RESPONSE_MAX_DELAY_MS = 1000;
-    private static final ThreadLocal<UsageDataSession> threadLocal = new InheritableThreadLocal<>();
+    private static final String AMBCTX_USAGE_DATA_SESSION = "UsageDataSession";
```
yaohaizh: Should avoid abbreviation for non-local variable/methods naming.

PR #121, opened 2017-11-29
Support send request in protocol server

testforstephen:

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/AbstractProtocolServer.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/AbstractProtocolServer.java
@@ -23,13 +23,21 @@ import java.io.Reader;
 import java.io.Writer;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.util.Timer;
+import java.util.TimerTask;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Consumer;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-public abstract class AbstractProtocolServer {
+import com.microsoft.java.debug.core.protocol.Events.DebugEvent;
+
+import io.reactivex.disposables.Disposable;
+import io.reactivex.subjects.PublishSubject;
+
+public abstract class AbstractProtocolServer implements IProtocolServer {
     private static final Logger logger = Logger.getLogger("java-debug");
     private static final int BUFFER_SIZE = 4096;
     private static final String TWO_CRLF = "\r\n\r\n";
@@ -44,6 +52,8 @@ public abstract class AbstractProtocolServer {
     private ByteBuffer rawData;
     private int contentLength = -1;
     private AtomicInteger sequenceNumber = new AtomicInteger(1);
+
+    private PublishSubject<Messages.Response> subject = PublishSubject.<Messages.Response>create();
 
     /**
      * Constructs a protocol server instance based on the given input stream and output stream.
@@ -86,7 +96,12 @@ public abstract class AbstractProtocolServer {
         this.terminateSession = true;
     }
 
-    protected void sendMessage(Messages.ProtocolMessage message) {
+    /**
+     * Send a request/response/event to the DA.
+     * @param message
+     *              the message.
+     */
+    private void sendMessage(Messages.ProtocolMessage message) {
         message.seq = this.sequenceNumber.getAndIncrement();
 
         String jsonMessage = JsonUtils.toJson(message);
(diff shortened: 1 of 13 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/IProtocolServer.java
```diff
@@ -0,0 +1,56 @@
...
+
+package com.microsoft.java.debug.core.protocol;
+
+import java.util.function.Consumer;
+
+import com.microsoft.java.debug.core.protocol.Events.DebugEvent;
+import com.microsoft.java.debug.core.protocol.Messages.Request;
+import com.microsoft.java.debug.core.protocol.Messages.Response;
+
+public interface IProtocolServer {
```
akaroml: Perfect! Thanks.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/AbstractProtocolServer.java
```diff
@@ -102,14 +117,80 @@ protected void sendMessage(Messages.ProtocolMessage message) {
...
+
+    @Override
+    public void sendRequest(Messages.Request request, Consumer<Messages.Response> cb) {
+        sendRequest(request, 0, cb);
+    }
+
+    @Override
+    public void sendRequest(Messages.Request request, long timeout, Consumer<Messages.Response> cb) {
+        Timer timer = new Timer();
+        Disposable[] disposable = new Disposable[1];
```
akaroml: Is this is a new trick? Not quite understand this one...

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/AbstractProtocolServer.java
```diff
@@ -102,14 +117,80 @@ protected void sendMessage(Messages.ProtocolMessage message) {
...
+        boolean[] executed = new boolean[]{false};
```
akaroml: And this one...

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/IProtocolServer.java
```diff
@@ -0,0 +1,56 @@
...
+     * Send a request to the DA. And create a timeout error response to the callback if no response is received at the give time.
+     *
+     * @param request
+     *            the request message.
+     * @param timeout
+     *            the maximum time (in millis) to wait.
+     * @param cb
+     *            the request call back function.
+     */
+    void sendRequest(Request request, long timeout, Consumer<Response> cb);
```
akaroml: One suggestion here. Can we use CompletableFuture<Response> as return value? This way we have the chance to also pass errors like a timeout exception.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/IProtocolServer.java
```diff
@@ -0,0 +1,56 @@
...
+public interface IProtocolServer {
+    /**
+     * Send a request to the DA.
+     *
+     * @param request
+     *            the request message.
+     * @param cb
+     *            the request call back function.
+     */
+    void sendRequest(Request request, Consumer<Response> cb);
```
akaroml: One suggestion here. Can we use CompletableFuture as return value? This way we have the chance to also pass errors like a timeout exception.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/AbstractProtocolServer.java
```diff
@@ -45,6 +53,8 @@
     private int contentLength = -1;
     private AtomicInteger sequenceNumber = new AtomicInteger(1);
 
+    private PublishSubject<Messages.Response> subject = PublishSubject.<Messages.Response>create();
```
akaroml: subject -> responseSubject

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/AbstractProtocolServer.java
```diff
@@ -102,14 +117,80 @@ protected void sendMessage(Messages.ProtocolMessage message) {
...
+    public void sendRequest(Messages.Request request, Consumer<Messages.Response> cb) {
+        sendRequest(request, 0, cb);
+    }
+
+    @Override
+    public void sendRequest(Messages.Request request, long timeout, Consumer<Messages.Response> cb) {
+        Timer timer = new Timer();
+        Disposable[] disposable = new Disposable[1];
+        boolean[] executed = new boolean[]{false};
+        disposable[0] = subject.filter(response -> response.request_seq == request.seq).subscribe((response) -> {
```
akaroml: If we only need to get one response from the subject, we can use take(1) and we don't need to dispose the subscription that way.
subject.filter(...).take(1).subscribe(...)

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/AbstractProtocolServer.java
```diff
@@ -102,14 +117,80 @@ protected void sendMessage(Messages.ProtocolMessage message) {
...
+            synchronized (executed) {
+                if (executed[0]) {
```
akaroml: By using take(1), this won't be called twice.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/AbstractProtocolServer.java
```diff
@@ -102,14 +117,80 @@ protected void sendMessage(Messages.ProtocolMessage message) {
...
+                            if (executed[0]) {
+                                return;
+                            }
+                            executed[0] = true;
+                        }
+                        if (disposable[0] != null) {
+                            disposable[0].dispose();
+                            disposable[0] = null;
+                        }
+                        cb.accept(new Messages.Response(request.seq, request.command, false, "timeout"));
```
akaroml: Reporting timeout error using the normal execution flow is not recommended. Please switch to CompletableFuture and try to completeExceptionally with a TimeoutException.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/AbstractProtocolServer.java
testforstephen: when use Disposable directly in subscriber, the compile told me the local variable Disposable may not be initialized. So use array to workaround the issue.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/IProtocolServer.java
markaw67: No problem

PR #124, opened 2017-12-04
Add basic HCR support

yaohaizh:
Signed-off-by: Yaohai Zheng <yaozheng@microsoft.com>

@akaroml @andxu @testforstephen @ansyral


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AttachRequestHandler.java
```diff
@@ -85,7 +86,9 @@
             options.put(Constants.PROJECTNAME, attachArguments.projectName);
         }
         ISourceLookUpProvider sourceProvider = context.getProvider(ISourceLookUpProvider.class);
-        sourceProvider.initialize(context.getDebugSession(), options);
+        sourceProvider.initialize(context, options);
```
akaroml: @testforstephen added a new handler that does post-processing after launching/attaching. Please consider putting the code there.

yaohaizh: Post launching is not implemented yet. That PR was closed.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
```diff
@@ -66,6 +66,7 @@
     // AutoCloseable
     @Override
     public void close() throws Exception {
+        requests.clear();
```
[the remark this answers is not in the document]
yaohaizh: Fixed.

yaohaizh:
Split into minor changes. The main class is JavaHotCodeReplaceProvider and other codes are clue.

andxu:
Please update it

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
```diff
@@ -70,6 +70,7 @@ public void close() throws Exception {
         subscriptions().forEach(subscription -> {
             subscription.dispose();
         });
```
andxu: why not to clear subscriptions?

andxu: why not to clear subscriptions similarly?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/StackFrameUtility.java
```diff
@@ -0,0 +1,72 @@
...
+*******************************************************************************/
+
+package com.microsoft.java.debug.core;
+
+import com.sun.jdi.AbsentInformationException;
+import com.sun.jdi.IncompatibleThreadStateException;
+import com.sun.jdi.ReferenceType;
+import com.sun.jdi.StackFrame;
+
+public final class StackFrameUtility {
```
andxu: StackFrameUtils for the same name conventions

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/IHotCodeReplaceListener.java
```diff
@@ -0,0 +1,19 @@
...
+*     Microsoft Corporation - initial API and implementation
+*******************************************************************************/
+
+package com.microsoft.java.debug.core.adapter;
+
+import java.util.EventListener;
+
+public interface IHotCodeReplaceListener extends EventListener {
+
+    public void redefineClasses(IRedefineClassEvent event);
```
andxu: remove public since it is default public?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AttachRequestHandler.java
andxu: I think all providers should call initialize too

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
yaohaizh: This is a reinstall case. Subscriptions doesn't changed.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/StackFrameUtility.java
yaohaizh: Actually  we have DebugUtility here, not use where is *Utils coming from.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/IHotCodeReplaceListener.java
yaohaizh: Removed.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,328 @@
...
+        public HotCodeReplaceEvent(String message) {
+            super("hotCodeReplace");
+            this.message = message;
+        }
+    }
+
+    @Override
+    public void initialize(IDebugAdapterContext context, Map<String, Object> options) {
+        // Listen to the built file events.
+        ResourcesPlugin.getWorkspace().addResourceChangeListener(this, IResourceChangeEvent.POST_BUILD);
```
andxu: where is the logic to remove this listener?

yaohaizh: NO place to remove this since there is no place for removing yet.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,328 @@
...
+                        }
+                        if (CLASS_FILE_EXTENSION.equals(resource.getFullPath().getFileExtension())) {
+                            IPath localLocation = resource.getLocation();
+                            if (localLocation != null) {
+                                String path = localLocation.toOSString();
+                                IClassFileReader reader = ToolFactory.createDefaultClassFileReader(path,
+                                        IClassFileReader.CLASSFILE_ATTRIBUTES);
+                                if (reader != null) {
+                                    // this name is slash-delimited
+                                    String qualifiedName = new String(reader.getClassName());
```
andxu: why new a string instance? string is immutable

andxu: I think we should reuse the source name info from .class file since fqn is not mapped one-to-one to the java file.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,328 @@
...
+                        name = sourceAttribute;
+                    }
+                    element = project.findElement(new Path(name));
+                }
+                if (element instanceof ICompilationUnit) {
+                    ICompilationUnit cu = (ICompilationUnit) element;
+                    return cu.getCorrespondingResource();
+                }
+            } catch (CoreException e) {
+                logger.log(Level.INFO, "Failed to get exception" + e.getMessage(), e);
```
andxu: the error message is not a complete cause.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
yaohaizh: reader.getClassName is a char[]

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/IRedefineClassEvent.java
```diff
@@ -0,0 +1,20 @@
...
+*     Microsoft Corporation - initial API and implementation
+*******************************************************************************/
+
+package com.microsoft.java.debug.core.adapter;
+
+import java.util.List;
+
+public interface IRedefineClassEvent {
+
+    List<String> getRedefinedClassNames();
```
akaroml: getClassNames

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/IProvider.java
```diff
@@ -13,16 +13,15 @@
...
-     * @param debugSession
-     *            The associated debug session
+     *
+     * @param debugContext
+     *            The associated debug context
      * @param options
      *            the options
      */
-    default void initialize(IDebugSession debugSession, Map<String, Object> options) {
+    default void initialize(IDebugAdapterContext debugContext, Map<String, Object> options) {
```
akaroml: The initialize method was introduced a month ago. But I didn’t find a complete story of using it properly. Seems like there is no certain chance for initializing a provider. Currently it can be called after launching or attaching. Do we really need to maintain a model inside a provider? Can a provider only consists of static functions?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
```diff
@@ -135,4 +156,26 @@
         return breakpoints;
     }
 
+    @Override
+    public void redefineClasses(IRedefineClassEvent event) {
+        List<String> typenames = event.getRedefinedClassNames();
+        if (typenames == null || typenames.isEmpty()) {
+            return;
+        }
+        IBreakpoint[] copy = manager.getBreakpoints().clone();
```
akaroml: Why do we need to clone this?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,328 @@
...
+ * The hot code replace provider listens for changes to class files and notifies
+ * the running debug session of the changes.
+ */
+public class JavaHotCodeReplaceProvider implements IHotCodeReplaceProvider, IResourceChangeListener {
+
+    private static final Logger logger = Logger.getLogger(Configuration.LOGGER_NAME);
+
+    private static final String CLASS_FILE_EXTENSION = "class"; //$NON-NLS-1$
+
+    private IDebugSession currentDebugSession;
```
akaroml: I didn’t see a use case of this member.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,328 @@
...
+
+    private IDebugAdapterContext context;
```
akaroml: Didn’t find the use case for this one either. Is it really necessary?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
```diff
@@ -35,16 +41,31 @@
 import com.microsoft.java.debug.core.protocol.Responses;
 import com.microsoft.java.debug.core.protocol.Types;
 
-public class SetBreakpointsRequestHandler implements IDebugRequestHandler {
+public class SetBreakpointsRequestHandler implements IDebugRequestHandler, IHotCodeReplaceListener {
```
akaroml: I think the HotCodeReplaceProvider should listen to code change events and perform whatever operation outside the debugger core. It can access the BreakpointManager to manipulate the breakpoints. This can be written outside this handler.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
yaohaizh: Because some code are removed in this PR. Would keep these for next PR.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
yaohaizh: For main logic. Some code truncate in this PR.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
yaohaizh: Current the breatpointmanager is internally to the core, we can expose it outside via context, but manipulate the breapionts in the HCR will break the encapsulation of the code.  HCR should only care about the HCR work and let other parts do corresponding post work to split the logic.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/IProvider.java
yaohaizh: If the provide (HCR provider) need some operation over the virtualmachine/threads/callstacks, it need access the context object.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/StackFrameUtility.java
```diff
@@ -0,0 +1,72 @@
+/*******************************************************************************
+* Copyright (c) 2017 Microsoft Corporation and others.
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Eclipse Public License v1.0
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v10.html
+*
+* Contributors:
+*     Microsoft Corporation - initial API and implementation
```
andxu: Please combine origin license from jdt code.
 
 *   IBM Corporation - initial API and implementation
 *     Yevgen Kogan - Bug 403475 - Hot Code Replace drops too much frames in some cases

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
andxu: Please mark a TODO comment of removing this change listener

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
testforstephen: +1
`subscriptions.clear()`

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/IHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,19 @@
...
+ *     Microsoft Corporation - initial API and implementation
+ *******************************************************************************/
+
+package com.microsoft.java.debug.core.adapter;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public interface IHotCodeReplaceProvider extends IProvider {
+    CompletableFuture<List<String>> completed();
```
akaroml: The name is a bit confusing. Suggest something like `applyCodeChanges`.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AttachRequestHandler.java
```diff
@@ -86,7 +87,9 @@
...
+        IHotCodeReplaceProvider hcrProvider = context.getProvider(IHotCodeReplaceProvider.class);
```
akaroml: The provider should be initialized whenever the `DebugAdapterContext` is ready. Here's the flow of provider:
1. A provider is only called by handlers, but not the opposite
2. A provider can be initialized right after the `DebugAdapterContext` is ready, and it can save the reference for later use.
3. Whenever there is a request from the handlers, the provider can then use the context.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
```diff
@@ -36,15 +40,21 @@
...
+    private IDebugAdapterContext context = null;
+
     @Override
     public List<Command> getTargetCommands() {
         return Arrays.asList(Command.SETBREAKPOINTS);
     }
 
     @Override
     public CompletableFuture<Response> handle(Command command, Arguments arguments, Response response, IDebugAdapterContext context) {
+        this.context = context;
```
akaroml: Instead of saving the context as a member, suggest making it an argument of `reinstallBreakpoints` in a more functional way.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AttachRequestHandler.java
testforstephen: Agree with @andxu, the context should have some interface to get all providers, then just iterate all providers and initialize them.

testforstephen: The DebugAdapterContext should return a ProviderRegistry instead of a concrete getProvider method. Then invoke ProviderRegistry to iterate all registered providers.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,343 @@
...
+                    needHotCodeReplace.wait();
+                }
+            } catch (InterruptedException e) {
+                logger.log(Level.INFO, "The hotcode replace completion was interrupted by " + e.getMessage(), e);
+            }
+        }
+        return CompletableFuture.completedFuture(new ArrayList<String>());
+    }
+
+    private void doHotCodeReplace(List<IResource> resources, List<String> fullyQualifiedName) {
```
akaroml: This is not fully implemented, right?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,343 @@
...
+                        needHotCodeReplace.set(false);
+                        needHotCodeReplace.notifyAll();
+                    }
+                }
+            }
+        }
+    }
+
+    @Override
+    public CompletableFuture<List<String>> completed() {
```
akaroml: This is not fully implemented, right?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/IProvider.java
testforstephen: The second parameter `options` is not necessary because debugContext contain all common information.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,343 @@
...
+    class ChangedClassFilesVisitor implements IResourceDeltaVisitor {
+        /**
+         * The collection of changed class files.
+         */
+        protected List<IResource> changedFiles = null;
+
+        /**
+         * Collection of qualified type names, corresponding to class files.
+         */
+        protected List<String> fullyQualifiedNames = null;
```
testforstephen: Where are initialize changedFile and fullyQualifiedNames initialized?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,343 @@
...
+                }
+            }
+        }
+    }
+
+    @Override
+    public CompletableFuture<List<String>> completed() {
+        synchronized (needHotCodeReplace) {
+            try {
+                needHotCodeReplace.wait(100);
```
testforstephen: what's the purpose for wait twice?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
```diff
@@ -77,6 +87,12 @@
             return AdapterUtils.createAsyncErrorResponse(response, ErrorCode.SET_BREAKPOINT_FAILURE,
                         String.format("Failed to setBreakpoint. Reason: '%s' is an invalid path.", bpArguments.source.path));
         }
+
+        if (bpArguments.sourceModified) {
+            IHotCodeReplaceProvider hcrProvider = context.getProvider(IHotCodeReplaceProvider.class);
+            hcrProvider.completed().thenAccept((List<String> result) -> reinstallBreakpoints(result));
```
testforstephen: for sourceModified case, the BreakpointManager actually will reinstallBreakpoint, is there any possible to cause race condition?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/IProvider.java
yaohaizh: I would leave the IProvider refactor later. Currently, the IProvider/IProviderContext/IDebugAdapterContext is in a chaos. 
I would like to take another iteration of refactor beside this changes.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AttachRequestHandler.java
yaohaizh: Just leave the IProvider/IProvdierContext refactor later and outside the scope of this change. We can have another PR for how to organize/initialize/uninstall IProviders.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
yaohaizh: TBA in another change

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,343 @@
...
+         * if no source file could be found.
+         *
+         * @param project
+         *            the java project containing the classfile
+         * @param qualifiedName
+         *            fully qualified name of the type, slash delimited
+         * @param sourceAttribute
+         *            debug source attribute, or <code>null</code> if none
+         */
+        private IResource getSourceFile(IJavaProject project, String qualifiedName, String sourceAttribute) {
```
testforstephen: Looking at the underlying implementation `JavaProject#findElement(IPath)`, it will compute the fully qualified name from path and then using fully qualified name to findType.

So here you can use `JavaProject#findType(fullyQualifiedName)` directly.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AttachRequestHandler.java
yaohaizh: Same as above

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
yaohaizh: yep, what's why i add clone in reinstallBreakpoints previously

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
yaohaizh: For functional's parameter, the parameter capture the local variable in its body. But as for the context, it is a global variable and doesn't change in this functional calling lifecycle.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
yaohaizh: This is the JDT's code. Leave it as-is since it is fully tested.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
yaohaizh: Long story. And it is the price we have to pay if use in a functional programming.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
yaohaizh: Yes, add TODO

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
yaohaizh: Should be in later 2nd part.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
testforstephen: I don't think JDT's code is right, getSourceFile and findElement in last line are unnecessary. Using findType can get the result directly.

If you keep the original implementation, suggest to put it in additional utility class so that we can recognize what's our own logic.

yaohaizh: This file is JDT implementation. See the license header. We only add glue code in this file.

yaohaizh: Since the JDT code is well tested, I'll trust their implementation here. There might be some corner case if JavaProject#findType fail  that we don't know. Nothing is guarantee until we do fully verification on the code.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,344 @@
...
+     * @param qualifiedTypeName
+     *            fully qualified type name
+     * @param project
+     *            project to search in
+     * @return class file or compilation unit or <code>null</code>
+     * @throws CoreException
+     *             if an exception occurs
+     */
+    public static IJavaElement findElement(String qualifiedTypeName, IJavaProject project) throws CoreException {
+        String path = qualifiedTypeName;
```
testforstephen: Use below logic to implement finding element for fqn.

<pre>
        // lookup type
        NameLookup lookup = newNameLookup(DefaultWorkingCopyOwner.PRIMARY);
        NameLookup.Answer answer = lookup.findType(
                qualifiedTypeName,
            false,
            NameLookup.ACCEPT_ALL,
            true/* consider secondary types */,
            false/* do NOT wait for indexes */,
            false/*don't check restrictions*/,
            null);

        if (answer != null) {
            return answer.type.getParent();
        } else {
            return null;
        }
</pre>

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
yaohaizh: Fixed

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
```diff
@@ -135,4 +148,25 @@
         return breakpoints;
     }
 
+    private void reinstallBreakpoints(IDebugAdapterContext context, List<String> typenames) {
+        if (typenames == null || typenames.isEmpty()) {
+            return;
+        }
+        // Avoid race condition here.
+        IBreakpoint[] breakpoints = manager.getBreakpoints().clone();
```
akaroml: Seems no race condition is here...

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/IHotCodeReplaceProvider.java
```diff
@@ -0,0 +1,19 @@
...
+ *     Microsoft Corporation - initial API and implementation
+ *******************************************************************************/
+
+package com.microsoft.java.debug.core.adapter;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+public interface IHotCodeReplaceProvider extends IProvider {
+    CompletableFuture<List<String>> redefinedClasses();
```
akaroml: redefineClasses

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
```diff
@@ -77,6 +84,12 @@
             return AdapterUtils.createAsyncErrorResponse(response, ErrorCode.SET_BREAKPOINT_FAILURE,
                         String.format("Failed to setBreakpoint. Reason: '%s' is an invalid path.", bpArguments.source.path));
         }
+
+        if (bpArguments.sourceModified) {
+            IHotCodeReplaceProvider hcrProvider = context.getProvider(IHotCodeReplaceProvider.class);
+            hcrProvider.redefinedClasses().thenAccept((List<String> result) -> reinstallBreakpoints(context, result));
```
akaroml: Seems like a blocking sync operation here.
@testforstephen suggested thenAcceptAsync as replacement.

yaohaizh:
Merge from master and commit the approved changes.

Issue #31, opened 2017-09-19
Should avoid to use SearchEngine in jdt which is not compatible with project settings.


andxu:
SearchEngine will resolve classes that is not in current project among workspace which makes the source mapping not compatible with project settings.

testforstephen:
duplicate with https://github.com/Microsoft/vscode-java-debug/issues/121, will discuss the issue there.

PR #127, opened 2017-12-07
Looking up the stack frame's associated source file from source containers directly to improve searching perf and bugs

testforstephen:
See the feature description https://github.com/Microsoft/vscode-java-debug/issues/121

Previous debugger use jdt search engine to look up the associated source file for the stack frame, but it has several issues:
- Search Engine will resolve classes that is not in current project among workspace which makes the source mapping not compatible with project settings. (Microsoft/java-debug#31)
- Search Engine cannot handle the class name with $ very well. (https://github.com/Microsoft/vscode-java-debug/issues/19)
- The Search policy is a kind of range search, and the result is not accurate. It could get multiple similar results for the search key so that it shows the wrong source for the stack trace. (https://github.com/Microsoft/vscode-java-debug/issues/101)
- The jdt search engine is based on index, for big project, it will take a very long time (even minutes) to construct the index file. So when import a new project in VSCode and start to debug it, the first stack frame call would take a little long time. But if the index is built, the search engine perf is acceptable.


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtUtils.java
```diff
@@ -78,4 +98,164 @@ public static IJavaProject getJavaProject(String projectName) {
...
+        List<IProject> projects = new ArrayList<>();
+
+        IProject targetProject = JdtUtils.getProject(currentProject);
+        if (targetProject != null) {
+            projects.add(targetProject);
+        }
+
+        List<IProject> workspaceProjects = Arrays.asList(ResourcesPlugin.getWorkspace().getRoot().getProjects());
+        projects.addAll(workspaceProjects.stream().filter((project) -> {
+            return !project.equals(targetProject);
```
akaroml: Maybe we can just use whatever returned by getProjects() instead of filtering out the targetProject. The logic seems redundant. The project collection should already have the target project. So this function can have no parameter.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtUtils.java
```diff
@@ -78,4 +98,164 @@ public static IJavaProject getJavaProject(String projectName) {
...
+                // Add source containers associated with the project's source folders.
+                containers.add(new JavaProjectSourceContainer(javaProject));
+            }
+        }
+
+        return containers.toArray(new ISourceContainer[0]);
+    }
+
+    private static ISourceContainer[] getSourceContainers(IJavaProject project, Set<IRuntimeClasspathEntry> calculated) {
+        if (project != null && project.exists()) {
```
akaroml: Use early return.

if (project == null || !project.exists()) {
     return *empty*;
}

In this way we avoid indenting large chunk of code.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -164,17 +170,32 @@ public boolean supportsRealtimeBreakpointVerification() {
 
     @Override
     public String getSourceFileURI(String fullyQualifiedName, String sourcePath) {
```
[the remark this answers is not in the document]
testforstephen: i'd like to change the interface to `public String getSourceFileURI(StackFrame stackFrame)`.
From the stack frame, the provider could get more info for accurate searching, such as class name, source name, source path, even source info for the specified stratum (like scala).

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtUtils.java
testforstephen: nice.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -164,17 +170,32 @@ public boolean supportsRealtimeBreakpointVerification() {
...
-        if (fullyQualifiedName == null) {
-            throw new IllegalArgumentException("fullyQualifiedName is null");
+        if (sourcePath == null) {
+            return null;
         }
-        // Jdt Search Engine doesn't support searching anonymous class or local type directly.
-        // But because the inner class and anonymous class have the same java file as the enclosing type,
-        // search their enclosing type instead.
-        if (fullyQualifiedName.indexOf("$") >= 0) {
-            return searchDeclarationFileByFqn(AdapterUtils.parseEnclosingType(fullyQualifiedName));
```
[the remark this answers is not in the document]
testforstephen: nice.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtUtils.java
testforstephen: The original purpose is `"If the project name is specified, it will put the source containers from the specified project in the front of the result. "`

The filter logic is to remove duplicated project to avoid duplicated computation.

Yes, it's unnecessary, because the final source container result is a set object, it'll remove the duplicated source container. The following code is OK.
<pre>
        IProject targetProject = JdtUtils.getProject(projectName);
        if (targetProject != null) {
            projects.add(targetProject);
        }

        List<IProject> workspaceProjects = Arrays.asList(ResourcesPlugin.getWorkspace().getRoot().getProjects());
        projects.addAll(workspaceProjects);
</pre>

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtUtils.java
```diff
@@ -78,4 +98,164 @@ public static IJavaProject getJavaProject(String projectName) {
...
+     * then the other projects at the same workspace.
+     * </p>
+     * <p>
+     * Otherwise, loop every projects at the current workspace and combine the parsed source containers directly.
+     * </p>
+     * @param currentProject
+     *                  the current project name.
+     * @return the possible source container list.
+     */
+    public static ISourceContainer[] getSourceContainers(String currentProject) {
```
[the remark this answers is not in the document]
testforstephen: nice.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtUtils.java
```diff
@@ -112,31 +111,29 @@ public static IProject getProject(String projectName) {
...
+        IProject targetProject = JdtUtils.getProject(projectName);
         if (targetProject != null) {
             projects.add(targetProject);
         }
 
         List<IProject> workspaceProjects = Arrays.asList(ResourcesPlugin.getWorkspace().getRoot().getProjects());
-        projects.addAll(workspaceProjects.stream().filter((project) -> {
-            return !project.equals(targetProject);
-        }).collect(Collectors.toList()));
+        projects.addAll(workspaceProjects);
```
akaroml: There can be duplicated entries inside the `projects` collection. The `targetProject` is already inside.
```
ResourcesPlugin.getWorkspace().getRoot().getProjects()
JdtUtils.getProject(projectName)
```
Does the return value of the first call contain the return value of the second call?

testforstephen: YES, it contains the second result.

testforstephen: But it's OK to have duplicated projects because the later computation code has logic to avoid repeated calculation on the duplicated classpath entries.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtUtils.java
```diff
@@ -78,4 +97,165 @@ public static IJavaProject getJavaProject(String projectName) {
...
+
+        IProject targetProject = JdtUtils.getProject(projectName);
+        if (targetProject != null) {
+            projects.add(targetProject);
+        }
+
+        List<IProject> workspaceProjects = Arrays.asList(ResourcesPlugin.getWorkspace().getRoot().getProjects());
+        projects.addAll(workspaceProjects);
+
+        Set<IRuntimeClasspathEntry> calculated = new LinkedHashSet<>();
```
yaohaizh: Use Java.Stream to optimize the coding logic

testforstephen: nice. fixed.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtUtils.java
```diff
@@ -78,4 +97,165 @@ public static IJavaProject getJavaProject(String projectName) {
...
+
+        projects.stream().map(project -> JdtUtils.getJavaProject(project))
```
akaroml: Suggest using distinct() to avoid duplicates.

testforstephen: fixed.

PR #129, opened 2017-12-11
fix unsupported breakpoint at method entry/exit or class

andxu:
Add location type check to avoid breakpoint on method/field which will never be hit in current implementation.


andxu:
https://github.com/Microsoft/vscode-java-debug/issues/90

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
```diff
@@ -80,17 +84,25 @@
         try {
             List<Types.Breakpoint> res = new ArrayList<>();
             IBreakpoint[] toAdds = this.convertClientBreakpointsToDebugger(sourcePath, bpArguments.breakpoints, context);
+            Set<IBreakpoint> processedBreakpoints = new HashSet<>();
             // See the VSCode bug https://github.com/Microsoft/vscode/issues/36471.
             // The source uri sometimes is encoded by VSCode, the debugger will decode it to keep the uri consistent.
             IBreakpoint[] added = manager.setBreakpoints(AdapterUtils.decodeURIComponent(sourcePath), toAdds, bpArguments.sourceModified);
             for (int i = 0; i < bpArguments.breakpoints.length; i++) {
+                // check whether the breakpoint has already been processed, otherwise the same breakpoint might be installed more than once.
+                // if class name is null, then this breakpoint is not valid, we doesn't need to install it or update hit count for it.
```
testforstephen: we doesn't -> we don't

akaroml:
I'm not sure about the purpose of this pull request. Let's sync more tomorrow.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -150,7 +150,10 @@ public boolean supportsRealtimeBreakpointVerification() {
                 ValidBreakpointLocationLocator locator = new ValidBreakpointLocationLocator(astUnit, lines[i], true, true);
                 astUnit.accept(locator);
                 // When the final valid line location is same as the original line, that represents it's a valid breakpoint.
-                if (lines[i] == locator.getLineLocation()) {
+                // Add location type check to avoid breakpoint on method/field which will never be hit in current implementation.
+                if (locator.getLocationType() == ValidBreakpointLocationLocator.LOCATION_LINE) {
```
testforstephen: Update the javadoc for `getFullyQualifiedName()` because the behavior is changed.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
```diff
@@ -105,17 +109,25 @@ public void initialize(IDebugAdapterContext context) {
...
             List<Types.Breakpoint> res = new ArrayList<>();
             IBreakpoint[] toAdds = this.convertClientBreakpointsToDebugger(sourcePath, bpArguments.breakpoints, context);
+            Set<IBreakpoint> processedBreakpoints = new HashSet<>();
             // See the VSCode bug https://github.com/Microsoft/vscode/issues/36471.
             // The source uri sometimes is encoded by VSCode, the debugger will decode it to keep the uri consistent.
             IBreakpoint[] added = manager.setBreakpoints(AdapterUtils.decodeURIComponent(sourcePath), toAdds, bpArguments.sourceModified);
             for (int i = 0; i < bpArguments.breakpoints.length; i++) {
+                // check whether the breakpoint has already been processed, otherwise the same breakpoint might be installed more than once.
+                // if class name is null, then this breakpoint is not valid, we don't need to install it or update hit count for it.
+                if (processedBreakpoints.contains(toAdds[i]) || added[i].className() == null) {
```
akaroml: BreakpointManager checks for duplicate breakpoints. Why do we need to check again?

andxu: eg: the following code 
    1.  // a
    2. //  b
    3. c++;

if we set all breakpoints for the 3 lines, manager.setBreakpoints will return an 3-element array all with the same breakpoint line # 3, when process the first breakpoint on line 1, actually we are installing breakpoint  on line 3, so the line 2, and line 3 breakpoints will not be processed again.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
andxu: updated

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java
```diff
@@ -88,7 +90,7 @@ public BreakpointManager() {
         }
 
         // Compute the breakpoints that are newly added.
-        List<IBreakpoint> toAdd = new ArrayList<>();
+        Set<IBreakpoint> toAdd = new HashSet<>();
```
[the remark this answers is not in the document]
andxu: in previous code, the breakpoint line will never be changed, invalid breakpoint will be marked with null class name, thus the breakpoint in toAdd  will never be the same, while in new design, the line number may be changed, eg:

1. // a
2. // b
3. c++;

if we use ArrayList, addBreakpointsInternally in the following code will add three same breakpints to this.breakpoints, which will bring duplicate ones in this.breakpoints

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
akaroml: In line 143, we have a todo item that indicates future support of moving breakpoint location to a valid one. Please remove the comments if you just implemented this.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -150,7 +150,10 @@ public boolean supportsRealtimeBreakpointVerification() {
...
+                    // update lines to the next line which is valid for breakpoint
+                    lines[i] = locator.getLineLocation();
```
akaroml: Modifying what's passed in is not recommended. Please add an override to accept an extra parameter for the modified line numbers.

akaroml: As discussed, we only support line breakpoint for now.

PR #148, opened 2018-01-22
Fix general exception for native methods.

yaohaizh:
@akaroml @testforstephen @andxu 

For spring case, it will try to pop stack above native methods but it will failed. It is the design behavior of JVM and similar behavior in eclipse. 
However, the exception is https://docs.oracle.com/javase/7/docs/jdk/api/jpda/jdi/com/sun/jdi/InternalException.html instead of more specific https://docs.oracle.com/javase/7/docs/jdk/api/jpda/jdi/com/sun/jdi/NativeMethodException.html, just try the general runtime exception as eclipse did. 

We need document the non support scenarios here.

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/StackFrameUtility.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/StackFrameUtility.java
@@ -14,6 +14,7 @@ package com.microsoft.java.debug.core;
 import com.sun.jdi.AbsentInformationException;
 import com.sun.jdi.IncompatibleThreadStateException;
 import com.sun.jdi.InvalidStackFrameException;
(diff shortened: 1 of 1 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/StackFrameUtility.java
```diff
@@ -32,7 +31,7 @@ public static boolean isNative(StackFrame frame) {
     public static void pop(StackFrame frame) throws DebugException {
         try {
             frame.thread().popFrames(frame);
-        } catch (IncompatibleThreadStateException | InvalidStackFrameException e) {
+        } catch (IncompatibleThreadStateException | RuntimeException e) {
```
akaroml: Add branches to catch more specific exception types as discussed.

PR #151, opened 2018-03-09
add condition field for conditional bp

andxu:


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/IDebugSession.java
```diff
@@ -32,6 +32,8 @@
 
     IBreakpoint createBreakpoint(String className, int lineNumber, int hitCount);
 
+    IBreakpoint createBreakpoint(String className, int lineNumber, int hitCount, String condition);
```
yaohaizh: For the interface, just keep the interface concise. If I understand correctly, only this new one is enough?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/IBreakpoint.java
```diff
@@ -27,4 +27,6 @@
     void putProperty(Object key, Object value);
 
     Object getProperty(Object key);
+
+    String condition();
```
yaohaizh: Update the naming convention consistently.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/IDebugSession.java
andxu: I will remove the other two unused apis.

PR #154, opened 2018-03-12
PR2 of conditional BP 

andxu:
1. Fix the issue when the condition is updated
2. Add a cache to the compiled expression for conditional breakpoint.

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java
@@ -32,7 +32,7 @@ public class BreakpointManager {
     private HashMap<String, HashMap<String, IBreakpoint>> sourceToBreakpoints;
     private AtomicInteger nextBreakpointId = new AtomicInteger(1);
     // BreakpointManager is the owner class of the breakpoint to compiled expression map, it will remove
-    // the breakpoint from this map if the breakpoint is removed or changing its condition
+    // the breakpoint from this map if the breakpoint is removed or its condition is changed
     private Map<IBreakpoint, Object> breakpointExpressionMap = new HashMap<>();
 
     /**
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java
@@ -31,6 +31,9 @@ public class BreakpointManager {
     private List<IBreakpoint> breakpoints;
     private HashMap<String, HashMap<String, IBreakpoint>> sourceToBreakpoints;
     private AtomicInteger nextBreakpointId = new AtomicInteger(1);
+    // BreakpointManager is the owner class of the breakpoint to compiled expression map, it will remove
+    // the breakpoint from this map if the breakpoint is removed or changing its condition
+    private Map<IBreakpoint, Object> breakpointExpressionMap = new HashMap<>();
 
     /**
      * Constructor.
@@ -77,6 +80,7 @@ public class BreakpointManager {
                 } catch (Exception e) {
                     logger.log(Level.SEVERE, String.format("Remove breakpoint exception: %s", e.toString()), e);
                 }
+                breakpointExpressionMap.remove(bp);
                 this.breakpoints.remove(bp);
             }
             this.sourceToBreakpoints.put(source, null);
@@ -143,6 +147,7 @@ public class BreakpointManager {
                     // Destroy the breakpoint on the debugee VM.
                     breakpoint.close();
                     this.breakpoints.remove(breakpoint);
+                    breakpointExpressionMap.remove(breakpoint);
                     breakpointMap.remove(String.valueOf(breakpoint.getLineNumber()));
                 } catch (Exception e) {
                     logger.log(Level.SEVERE, String.format("Remove breakpoint exception: %s", e.toString()), e);
@@ -164,6 +169,28 @@ public class BreakpointManager {
             return new IBreakpoint[0];
         }
         return breakpointMap.values().toArray(new IBreakpoint[0]);
+    }
+
+
+    /**
+     * Get the compiled expression map with breakpoint, it will be used in JdtEvaluationProvider#evaluateForBreakpoint for storing
+     * the compiled expression when the first time this conditional breakpoint is hit.
+     *
+     * @return the compiled expression map
+     */
+    public Map<IBreakpoint, Object> getBreakpointExpressionMap() {
+        return breakpointExpressionMap;
+    }
+
+    /**
+     *  Update the condition for the specified breakpoint, and clear the compiled expression for the breakpoint.
+     *
+     * @param breakpoint the conditional breakpoint
+     * @param newCondition the new condition to be used.
+     */
+    public void updateConditionCompiledExpression(IBreakpoint breakpoint, String newCondition) {
+        breakpoint.setCondition(newCondition);
+        breakpointExpressionMap.remove(breakpoint);
     }
 
     /**
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
@@ -157,10 +157,7 @@ public class SetBreakpointsRequestHandler implements IDebugRequestHandler {
     private void registerBreakpointHandler(IDebugAdapterContext context) {
         IDebugSession debugSession = context.getDebugSession();
         if (debugSession != null) {
-            debugSession.getEventHub().events().subscribe(debugEvent -> {
-                if (!(debugEvent.event instanceof BreakpointEvent)) {
-                    return;
-                }
+            debugSession.getEventHub().events().filter(debugEvent -> debugEvent.event instanceof BreakpointEvent).subscribe(debugEvent -> {
                 Event event = debugEvent.event;
                 if (debugEvent.eventSet.size() > 1 && debugEvent.eventSet.stream().anyMatch(t -> t instanceof StepEvent)) {
                     // The StepEvent and BreakpointEvent are grouped in the same event set only if they occurs at the same location and in the same thread.
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
@@ -76,7 +76,6 @@ public class SetBreakpointsRequestHandler implements IDebugRequestHandler {
                     logger.severe(e.toString());
                 }
             });
-
     }
 
     @Override
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
@@ -15,6 +15,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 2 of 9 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java
```diff
@@ -31,6 +31,9 @@
     private List<IBreakpoint> breakpoints;
     private HashMap<String, HashMap<String, IBreakpoint>> sourceToBreakpoints;
     private AtomicInteger nextBreakpointId = new AtomicInteger(1);
+    // BreakpointManager is the owner class of the breakpoint to compiled expression map, it will remove
+    // the breakpoint from this map if the breakpoint is removed or changing its condition
```
yaohaizh: if the breakpoint is removed or changing its condition ==> if the breakpoint is removed or its condition is changed

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
```diff
@@ -131,9 +132,16 @@ public void initialize(IDebugAdapterContext context) {
...
-                    // Update hitCount condition.
-                    added[i].setHitCount(toAdds[i].getHitCount());
+                } else if (added[i].className() != null) {
+                    if (toAdds[i].getHitCount() != added[i].getHitCount()) {
+                        // Update hitCount condition.
+                        added[i].setHitCount(toAdds[i].getHitCount());
+                    }
+
+                    if (!StringUtils.equals(toAdds[i].getCondition(), added[i].getCondition())) {
+                        manager.updateConditionCompiledExpression(added[i], toAdds[i].getCondition());
```
yaohaizh: It is strange put the hitCount update in here while keep the condition update in other places. The logic is mixed.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
```diff
@@ -178,6 +188,37 @@ public void clearState(ThreadReference thread) {
...
+            if (project == null) {
+                if (StringUtils.isBlank(projectName)) {
+                    logger.severe("Cannot evaluate when project is not specified.");
+                    throw new IllegalStateException("Please specify projectName in launch.json.");
+                }
+                project = JdtUtils.getJavaProject(projectName);
+            }
+
+            if (project == null) {
+                new IllegalStateException(String.format("Project %s cannot be found.", projectName));
```
yaohaizh: Why new a unused exception?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
```diff
@@ -178,6 +188,37 @@ public void clearState(ThreadReference thread) {
...
+                return false;
+            }
+            if (launch == null) {
+                launch = createILaunchMock(project);
+            }
+        }
+
+        if (debugTarget == null) {
```
yaohaizh: The outside is already have this logic

```java
debugTarget == null
```

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
```diff
@@ -178,6 +188,37 @@ public void clearState(ThreadReference thread) {
         }
     }
 
+    private boolean ensureDebugTarget(VirtualMachine vm) {
+        String projectName = (String) options.get(Constants.PROJECTNAME);
```
yaohaizh: If debugTarget is not null, this line is not necessary ?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java
```diff
@@ -166,6 +171,28 @@ private void removeBreakpointsInternally(String source, IBreakpoint[] breakpoint
...
     }
 
+
+    /**
+     * Get the compiled expression map with breakpoint, it will be used in JdtEvaluationProvider#evaluateForBreakpoint for storing
+     * the compiled expression when the first time this conditional breakpoint is hit.
+     *
+     * @return the compiled expression map
+     */
+    public Map<IBreakpoint, Object> getBreakpointExpressionMap() {
```
yaohaizh: If the cache is only used by the evaluation Provider,  should it be better put insider the evaluation provider?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
```diff
@@ -68,68 +69,53 @@ public void initialize(IDebugAdapterContext context, Map<String, Object> props)
...
+
+        try  {
+            ensureDebugTarget(thread.virtualMachine());
+            JDIThread jdiThread = getMockJDIThread(thread);
+            JDIStackFrame stackframe = (JDIStackFrame) jdiThread.getTopStackFrame();
+
+            ASTEvaluationEngine engine = new ASTEvaluationEngine(project, debugTarget);
+            ICompiledExpression ie = (ICompiledExpression) breakpointExpressionMap
+                    .computeIfAbsent(breakpoint, bp -> engine.getCompiledExpression(bp.getCondition(), stackframe));
+
```
yaohaizh: Return failureFuture and internalEvaluate result are two different instance. The client cannot hoop up with both future.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java
```diff
@@ -31,6 +31,9 @@
     private List<IBreakpoint> breakpoints;
     private HashMap<String, HashMap<String, IBreakpoint>> sourceToBreakpoints;
     private AtomicInteger nextBreakpointId = new AtomicInteger(1);
+    // BreakpointManager is the owner class of the breakpoint to compiled expression map, it will remove
+    // the breakpoint from this map if the breakpoint is removed or its condition is changed
+    private Map<IBreakpoint, Object> breakpointExpressionMap = new HashMap<>();
```
akaroml: Why is the breakpoint manager maintaining the cache map but not the evaluation engine?

akaroml: The cache should be kept by evaluation engine. The BreakpointManger notifies the engine to invalidate cache entries.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
akaroml: Why do we have two if blocks with the same condition to check whether debugTarget is null?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
```diff
@@ -178,6 +188,37 @@ public void clearState(ThreadReference thread) {
         }
     }
 
+    private boolean ensureDebugTarget(VirtualMachine vm) {
```
akaroml: The return value of ensureDebugTarget is never referenced anywhere. In the implementation, you use both exception and return value to indicate success, which is not recommended.

'ensure' means make sure something is actually there. When the function is not able to ensure that, an exception should be thrown. So please only use exceptions and the return value should be void.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
```diff
@@ -68,68 +69,53 @@ public void initialize(IDebugAdapterContext context, Map<String, Object> props)
...
-            }
-
-            if (project == null) {
-                completableFuture.completeExceptionally(new IllegalStateException(String.format("Project %s cannot be found.", projectName)));
-                return completableFuture;
-            }
-            if (launch == null) {
-                launch = createILaunchMock(project);
-            }
+    public CompletableFuture<Value> evaluateForBreakpoint(IBreakpoint breakpoint, ThreadReference thread, Map<IBreakpoint, Object> breakpointExpressionMap) {
```
akaroml: What's the difference of evaluating and evaluating for breakpoints?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
andxu: Because updating condition will need to clear the expression cache.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java
andxu: Because evaluation engine has no information about when the breakpoint is removed or updated.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
andxu: Because conditional breakpoint will be executed multiple time inside a loop(eval and cont), and the condition and location don't change, we need to optimize the conditional breakpoint, consider the following case, the i == 100 will be evaluated 999 times before the breakpoint is to hit.

 for (; i < 1000 + 1; i++) {
            if (i == 1000)  { # conditional BP of i == 1000
                System.out.println( "1111" );        
            } 
        }

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
andxu: Old issue in previous code, will fix it.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
andxu: will use exception for the failure of initialization to  debugTarget

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
andxu: sure

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
andxu: sure

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
andxu: done with a CompletableFuture parameter in internalEvaluate

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java
andxu: the keys are managed by  breakpoint manager, and the values are managed by evaluation evaluation provider, I don't think its better for breakpoint manager to reference JdtEvaluationProvider

andxu: JDT are storing expression in JavaLineBreakpoint@fCompiledExpressions

	/**
	 * The map of cached compiled expressions (ICompiledExpression) for this
	 * breakpoint, keyed by thread. This value must be cleared every time the
	 * breakpoint is added to a target.
	 */
	private Map<IJavaThread, ICompiledExpression> fCompiledExpressions = new HashMap<>();

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
andxu: setHitCount contains the logic to update hit count in jvm, it is not a pure set function.

PR #155, opened 2018-03-13
notify user when error is raised when evaluating breakpoint condition

andxu:


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Events.java
```diff
@@ -194,4 +194,22 @@ public HotCodeReplaceEvent(ChangeType changeType, String message) {
             this.message = message;
         }
     }
+
+    public static class UserNotificationEvent extends DebugEvent {
+        public enum NotifyType {
+            ERROR, WARNING
```
yaohaizh: For notification type, there should be information only type.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Events.java
```diff
@@ -194,4 +194,22 @@ public HotCodeReplaceEvent(ChangeType changeType, String message) {
             this.message = message;
         }
     }
+
+    public static class UserNotificationEvent extends DebugEvent {
+        public enum NotifyType {
+            ERROR, WARNING, INFORMATION
+        }
+
+        public NotifyType notifyType;
```
yaohaizh: NotificationType

andxu: will submit later

andxu: changed at https://github.com/Microsoft/java-debug/pull/156

andxu:
since andy_dev4 is merged to master, close this one and create a new one instead

andxu:
using this one, since the andy_cond3 has conflicts with master https://github.com/Microsoft/java-debug/pull/158

PR #158, opened 2018-03-15
notify user when error is raised when evaluating breakpoint condition[1]

andxu:
since andy_dev4 is merged to master, close this one and create a new one instead


andxu:
https://github.com/Microsoft/java-debug/pull/155

PR #162, opened 2018-03-26
fix the invalid stack exception during evaluation.

andxu:
1, the LeftCurly rule has been deprecated.
2. fix the invalid stack exception during evaluation.


yaohaizh:
Please clearly define what the commit is about, but not adding link. People are not willing to click link when check the commit log.

andxu:
using https://github.com/Microsoft/java-debug/pull/163 instead

PR #163, opened 2018-03-26
fix the invalid stack exception during evaluation

andxu:


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
```diff
@@ -95,7 +95,7 @@ public void initialize(IDebugAdapterContext context, Map<String, Object> props)
         try  {
             ensureDebugTarget(thread.virtualMachine(), thread, 0);
             JDIThread jdiThread = getMockJDIThread(thread);
-            JDIStackFrame stackframe = (JDIStackFrame) jdiThread.getTopStackFrame();
+            JDIStackFrame stackframe = new JDIStackFrame(jdiThread, thread.frame(0), 0);
```
testforstephen: Why update it? Is `JDIStackFrame` changed by jdt library?

andxu: getTopStackFrame uses the inner cache in Jdithread, it will have the possibility to become invalid

PR #164, opened 2018-03-26
fix an error when main class is not specified, evaluation doesn't work.

andxu:


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
```diff
@@ -198,6 +198,7 @@ private void findJavaProjectByStackFrame(ThreadReference thread, int depth) {
             visitedClassNames.add(typeName);
             if (validProjects.size() == 1) {
                 project = validProjects.get(0);
+                return;
             } else if (validProjects.size() == 0) {
                 logger.severe("No project is available for evaluation.");
                 throw new IllegalStateException("No project is available for evaluation, .");
```
SummerSun: this exception and the below one, will just be ignored in the catch block, why we throw them

andxu: The error message is different.

PR #169, opened 2018-04-09
Add auto completion implementation for debug console.

yaohaizh:
@andxu @testforstephen @akaroml

### DIFF
--- a/com.microsoft.java.debug.plugin/META-INF/MANIFEST.MF
+++ b/com.microsoft.java.debug.plugin/META-INF/MANIFEST.MF
@@ -16,7 +16,8 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.jdt.ls.core,
  org.eclipse.jdt.launching,
  com.google.gson;bundle-version="2.7.0",
- org.apache.commons.lang3;bundle-version="3.1.0"
+ org.apache.commons.lang3;bundle-version="3.1.0",
+ org.eclipse.lsp4j;bundle-version="0.2.0"
 Bundle-ClassPath: lib/commons-io-2.5.jar,
  .,
  lib/rxjava-2.1.1.jar,
--- a/com.microsoft.java.debug.plugin/META-INF/MANIFEST.MF
+++ b/com.microsoft.java.debug.plugin/META-INF/MANIFEST.MF
@@ -15,9 +15,9 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.jdt.core,
  org.eclipse.jdt.ls.core,
  org.eclipse.jdt.launching,
- com.google.gson;bundle-version="2.7.0",
- org.apache.commons.lang3;bundle-version="3.1.0",
- org.eclipse.lsp4j;bundle-version="0.2.0"
+ com.google.gson,
+ org.apache.commons.lang3,
+ org.eclipse.lsp4j
 Bundle-ClassPath: lib/commons-io-2.5.jar,
  .,
  lib/rxjava-2.1.1.jar,
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/CompletionsProvider.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/CompletionsProvider.java
@@ -13,31 +13,35 @@ package com.microsoft.java.debug.plugin.internal;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jdt.core.CompletionProposal;
+import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IType;
-import org.eclipse.jdt.core.search.IJavaSearchConstants;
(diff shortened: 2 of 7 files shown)


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/CompletionsProvider.java
```diff
@@ -0,0 +1,118 @@
...
+    private static final Logger logger = Logger.getLogger(Configuration.LOGGER_NAME);
+
+    @Override
+    public List<CompletionItem> codeComplete(StackFrame frame, String snippet, int line, int column) {
+        List<CompletionItem> res = new ArrayList<CompletionItem>();
+
+        try {
+            IType type = resolveType(frame);
+            if (type != null) {
+                final int offset = JsonRpcHelpers.toOffset(type.getCompilationUnit().getBuffer(), frame.location().lineNumber(), 0);
```
testforstephen: In order to keep the least dependency on jdt.ls, previously we always try to avoid using jdt.ls code in our implementation. Is it possible to keep the completeProvider implementation in jdt.ls, and then send the RPC request to jdt.ls for result?

yaohaizh: The main implementation is different from the jdt.ls.  The JDT is code complete on compilation unit with offset without *snippet* while our implementation is based on *snippet*. 

Here we leverage the CompletionProposalRequestor from the JDT.LS. Otherwise, it just copy this class to the debugger side which is not necessary.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/CompletionsProvider.java
```diff
@@ -0,0 +1,118 @@
...
+            logger.log(Level.WARNING, String.format("Failed to code complete because of %s", e.toString()), e);
+        }
+
+        return res;
+    }
+
+    private IType resolveType(StackFrame frame) throws CoreException, DebugException {
+        String fullyQualifiedTypeName = JdtUtils.getDeclaringTypeName(frame);
+
+        SearchPattern pattern = SearchPattern.createPattern(fullyQualifiedTypeName, IJavaSearchConstants.TYPE, IJavaSearchConstants.DECLARATIONS,
```
testforstephen: Search engine doesn't support anonymous class. Did you try the case?

com.microsoft.java.debug.plugin/META-INF/MANIFEST.MF
```diff
@@ -16,7 +16,8 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.jdt.ls.core,
  org.eclipse.jdt.launching,
  com.google.gson;bundle-version="2.7.0",
- org.apache.commons.lang3;bundle-version="3.1.0"
+ org.apache.commons.lang3;bundle-version="3.1.0",
+ org.eclipse.lsp4j;bundle-version="0.2.0"
```
testforstephen: Remove the bundle version limitation.

testforstephen: `lsp4j` is a new dependency, can it be installed by `mvn install`?

yaohaizh: It is part of language server dependency, which is also available in plugin dependency.

testforstephen: The ci server will use `mvn package` to build the jar, so it can resolve the dependency correctly. Right?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/CompletionsProvider.java
yaohaizh: Fixed.

com.microsoft.java.debug.plugin/META-INF/MANIFEST.MF
yaohaizh: Fixed for version issue.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/CompletionsProvider.java
```diff
@@ -0,0 +1,105 @@
...
+            logger.log(Level.WARNING, String.format("Failed to code complete because of %s", e.toString()), e);
+        }
+
+        return res;
+    }
+
+    private IType resolveType(StackFrame frame) throws CoreException, DebugException {
+        ISourceLookUpProvider sourceProvider = context.getProvider(ISourceLookUpProvider.class);
+        if (sourceProvider instanceof JdtSourceLookUpProvider) {
+            IJavaProject project = JdtUtils.findProject(frame, ((JdtSourceLookUpProvider) sourceProvider).getSourceContainers());
```
testforstephen: Need consider the case where cannot find the project for the frame.

yaohaizh: Fixed.

Issue #145, opened 2018-01-09
Support build using Java 9 SDK


yaohaizh:
The tools.jar is changed in the Java9 SDK because module system.

andxu:
Jdt.ls already supports jdk 9/10

testforstephen:
The issue still exists at debug.core. It's about tools.jar dependency.

PR #178, opened 2018-05-23
Fix the project's build errors in JDK 9/10

testforstephen:

Fix the issue https://github.com/Microsoft/java-debug/issues/145.

### DIFF
--- a/com.microsoft.java.debug.core/pom.xml
+++ b/com.microsoft.java.debug.core/pom.xml
@@ -40,13 +40,6 @@
     </build>
     <dependencies>
         <dependency>
-            <groupId>com.sun</groupId>
-            <artifactId>tools</artifactId>
-            <version>1.8</version>
-            <scope>system</scope>
-            <systemPath>${java.home}/../lib/tools.jar</systemPath>
-        </dependency>
-        <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
             <version>3.6</version>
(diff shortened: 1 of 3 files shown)


com.microsoft.java.debug.core/pom.xml
```diff
@@ -85,4 +78,21 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <profiles>
+    <profile>
+            <id>default-tools.jar</id>
+            <activation>
+                <jdk>(,9)</jdk>
```
yaohaizh: 9 is not official supported by Oracle. Just start with 10 is okay.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/DebugUtility.java
```diff
@@ -118,6 +118,18 @@ public static IDebugSession launch(VirtualMachineManager vmManager,
         List<LaunchingConnector> connectors = vmManager.launchingConnectors();
         LaunchingConnector connector = connectors.get(0);
 
+        /** In the sun JDK 10, the first launching connector is com.sun.tools.jdi.RawCommandLineLauncher, which is not the one we want to use.
+         *  Add the logic to filter the right one from LaunchingConnector list.
+         *
+         *  This fix is only for the JDI implementation by JDK. Other JDI implementations (such as JDT) doesn't have the impact.
+         */
+        for (LaunchingConnector con : connectors) {
+            if (con.getClass().getName().equals("com.sun.tools.jdi.SunCommandLineLauncher")) {
```
yaohaizh: Make the string final static value as the other values.

testforstephen: Nice. Fixed.

PR #179, opened 2018-05-28
initial commit for remote deubug plugin part.

andxu:


com.microsoft.java.debug.plugin/plugin.xml
```diff
@@ -7,6 +7,10 @@
             <command id="vscode.java.resolveClasspath"/>
             <command id="vscode.java.resolveMainClass"/>
             <command id="vscode.java.buildWorkspace"/>
```
yaohaizh: IMO, remote debug session should make changes on the launch.json configuration, but not add another set of debug commands.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaDebugDelegateCommandHandler.java
```diff
@@ -32,6 +34,18 @@
 
     public static String UPDATE_DEBUG_SETTINGS = "vscode.java.updateDebugSettings";
 
+    public static String START_REMOTE_DEBUG_SESSION = "vscode.java.startRemoteDebugSession";
```
testforstephen: It's better not to mix the remote related code into the existing code? Move them into a separate class and package folder?

akaroml:
There's a typo in the title.
deubug -> debug

PR #192, opened 2018-07-30
Add script to push p2 artifact

testforstephen:

I didn't find any existing jenkins plugin to deploy to bintray, so add a script to push the p2 artifacts to bintray.

The index page https://dl.bintray.com/vscjavaci/javadebug/ will show the latest version:
image

The sub path https://dl.bintray.com/vscjavaci/javadebug/releases will list the history:
image

### DIFF
--- a/com.microsoft.java.debug.repository/pushToBintray.sh
+++ b/com.microsoft.java.debug.repository/pushToBintray.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+#Usage: ./pushToBintray.sh username apikey repo package
+BINTRAY_USER=$1
+BINTRAY_API_KEY=$2
+BINTRAY_REPO=$3
+PCK_NAME=$4
+
+SCRIPT=$(readlink -f "$0")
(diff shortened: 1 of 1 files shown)


yaohaizh:
Can we directly call maven task/plugin to deploy theses?

testforstephen:
@yaohaizh This is a good question. It pushes me to look more into the bintray integration. And the existing bintray maven plugins mainly take bintray as maven central (Just like Nexus Sonatype repository manager). If we are deploying our product as maven artifacts,  these maven plugins are definitely recommended. 

But our target is deploying as p2 artifacts, whose structure is a little different with maven artifacts, and most of maven plugins don't have enough support for p2.  The closest maven plugin i found is https://github.com/random-maven/bintray-maven-plugin, which has special support for p2 upload. The plugin supports uploading folder content, such as eclipse p2 repository, to fixed path. But the limitation is that it only allows to upload the p2 to one fixed path. In our case, we need upload the p2 artifacts to two places, one is the latest place, another is the history versions.  The REST API solution can help us to achieve that. I prefer to keep the solution.

PR #194, opened 2018-08-01
upgrade to latest eclipse version: photon

andxu:


pom.xml
```diff
@@ -149,10 +149,10 @@
         </pluginManagement>
     </build>
     <repositories>
-        <repository>
-            <id>oxygen</id>
+       <repository>
+            <id>photon</id>
             <layout>p2</layout>
-            <url>http://download.eclipse.org/releases/oxygen</url>
+            <url>http://download.eclipse.org/releases/photon</url>
```
testforstephen: why we need update to `photon`? what's the different?

andxu: we will need it later, since photon is the lts now.

PR #198, opened 2018-08-04
support to start without debugging

Eskibear:
See https://github.com/Microsoft/vscode-java-debug/issues/351

This PR identifies `noDebug` parameter in the `launch` request, and does following thing:
* for `threads`, respond with an empty list.
* for `setBreakpoints`, respond with `verified: false`.

nodebug

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapterContext.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapterContext.java
@@ -234,7 +234,7 @@ public class DebugAdapterContext implements IDebugAdapterContext {
     }
 
     @Override
-    public boolean isDebugging() {
+    public boolean isDebugMode() {
         return this.isDebugMode;
     }
 
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapterContext.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapterContext.java
@@ -38,7 +38,7 @@ public class DebugAdapterContext implements IDebugAdapterContext {
     private Charset debuggeeEncoding;
     private transient boolean vmTerminated;
     private boolean isVmStopOnEntry = false;
-    private boolean isDebugging = true;
+    private boolean isDebugMode = true;
     private String mainClass;
     private StepFilters stepFilters;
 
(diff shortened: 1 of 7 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/IDebugAdapterContext.java
```diff
@@ -101,4 +101,8 @@
     StepFilters getStepFilters();
 
     IStackFrameManager getStackFrameManager();
+
+    boolean isDebugging();
```
yaohaizh: For interface design, it should keep the order of APIs either in alphabetically or grouping by the meaning.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/SetBreakpointsRequestHandler.java
```diff
@@ -86,7 +86,15 @@ public void initialize(IDebugAdapterContext context) {
         if (context.getDebugSession() == null) {
             return AdapterUtils.createAsyncErrorResponse(response, ErrorCode.EMPTY_DEBUG_SESSION, "Empty debug session.");
         }
-
```
yaohaizh: 1. If don't handle these requests(breakpoint/threads, etcs), we should either not register handlers in the DebugAdapter or just pass a DummyRequestHandlers.. 
2. There are requests more than threads/breakpoints should also be handled.

testforstephen: In nodebug mode, VS Code Debug Adapter still send setBreakpoints request to the debugger, looks like a bug, can you confirm with their team?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapterContext.java
```diff
@@ -38,6 +38,7 @@
     private Charset debuggeeEncoding;
     private transient boolean vmTerminated;
     private boolean isVmStopOnEntry = false;
+    private boolean isDebugging = true;
```
testforstephen: I don't see you change the `LaunchRequestHandler` logic, please remove the debug options from launching command, too.

Eskibear:
It's strange for a debugger to handle "noDebug" things, and the feature should be implemented in vscode-java-debug. When "noDebug" is specified in launch configuration, we can spawn a local java process instead of starting a debug session to make things complicated. 

So close it.

PR #199, opened 2018-08-06
Filter out the stale Main Classes cached by jdt search engine

testforstephen:
One fix for the problem `Failed to resolve classpath`: https://github.com/Microsoft/vscode-java-debug/issues/331

In some scenario (e.g. the bug https://github.com/Microsoft/vscode-java-debug/issues/349), the searching result from the jdt search engine contains some out dated main classes which don't exist in the workspace. This PR will filter out this kind of staled main classes.


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainClassHandler.java
```diff
@@ -72,7 +72,7 @@ public void acceptSearchMatch(SearchMatch match) {
                     try {
                         if (method.isMainMethod()) {
                             IResource resource = method.getResource();
-                            if (resource != null) {
+                            if (resource != null && resource.exists()) {
```
yaohaizh: This will not work. Resource are Java Model object which can be existing even without physical file. You need check the file. See https://github.com/eclipse/eclipse.jdt.ls/commit/064a6a57ff1eb751bf3cc6279909a97e126f501a

PR #201, opened 2018-08-13
fix issue 311:lazy load object variables while attach debug

andxu:


andxu:
https://github.com/Microsoft/vscode-java-debug/issues/311

andxu:
The variable evaluation is quite quick on local network, it will not be required to change the expensive value here, for remote debug, we have provided another way of using remote debug agent, eg: https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-debugger-azds

PR #204, opened 2018-08-14
Disable unnecessary request handlers for noDebug mode.

Eskibear:
For debug mode, enable all the request handlers. For noDebug mode, only enable handlers for initialize, configurationDone, launch, disconnect requests.

Ref:
https://github.com/Microsoft/vscode-java-debug/issues/351#issuecomment-412730985

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
@@ -48,14 +48,16 @@ public class DebugAdapter implements IDebugAdapter {
     private static final Logger logger = Logger.getLogger(Configuration.LOGGER_NAME);
 
     private IDebugAdapterContext debugContext = null;
-    private Map<Command, List<IDebugRequestHandler>> requestHandlers = null;
+    private Map<Command, List<IDebugRequestHandler>> requestHandlersForDebug = null;
+    private Map<Command, List<IDebugRequestHandler>> requestHandlersForRun = null;
 
     /**
      * Constructor.
      */
     public DebugAdapter(IProtocolServer server, IProviderContext providerContext) {
         this.debugContext = new DebugAdapterContext(server, providerContext);
-        requestHandlers = new HashMap<>();
+        requestHandlersForDebug = new HashMap<>();
+        requestHandlersForRun = new HashMap<>();
         initialize();
     }
 
@@ -73,7 +75,8 @@ public class DebugAdapter implements IDebugAdapter {
             // the operation is meaningless
             return CompletableFuture.completedFuture(response);
         }
-        List<IDebugRequestHandler> handlers = requestHandlers.get(command);
+        List<IDebugRequestHandler> handlers = this.debugContext.isDebugMode()
+                ? requestHandlersForDebug.get(command) : requestHandlersForRun.get(command);
         if (handlers != null && !handlers.isEmpty()) {
             CompletableFuture<Messages.Response> future = CompletableFuture.completedFuture(response);
             for (IDebugRequestHandler handler : handlers) {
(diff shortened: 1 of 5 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
```diff
@@ -113,6 +116,19 @@ private void initialize() {
...
     private void registerHandler(IDebugRequestHandler handler) {
+        registerHandler(handler, true, false);
+    }
+
+    private void registerHandler(IDebugRequestHandler handler, boolean forDebug, boolean forRun) {
+        if (forDebug) {
+            registerHandler(requestHandlersForDebug, handler);
+        }
+        if (forRun) {
+            registerHandler(requestHandlersForRun, handler);
```
andxu: I don't think parameter forDebug forRun will help to understand the handlers, I suguest to have two registerHandlerForDebug and registerHandlerForNonDebug private helper methods.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapterContext.java
```diff
@@ -231,4 +232,14 @@ public StepFilters getStepFilters() {
...
         return stackFrameManager;
     }
+
+    @Override
+    public boolean isDebugMode() {
+        return this.isDebugMode;
+    }
+
+    @Override
+    public void setDebugMode(boolean isDebugMode) {
```
Eskibear: Name of this setter is also a little strange for me, any naming suggestion ?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
Eskibear: Agree. Naming is always a tough thing. Since the flag is named `noDebug`, here should I use `registerHandlerForNonDebug` or `registerHandlerForNoDebug`. @andxu @testforstephen

testforstephen: Keep the consistent. I like `registerHandlerforNoDebug`

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapterContext.java
testforstephen: Add an enum type `enum LaunchMode { debug, noDebug}`. Then the setter could be `setLaunchMode(LaunchMode mode)`

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
testforstephen: About naming, here is a good suggestion.

`
Pro Tip: don't try and name a function by looking at its implementation as you'll only come up with a name that reflects how it works. Instead look at the call sites and see what they want to know or do.
`

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapterContext.java
testforstephen: Words from Effective Java:
> Prefer two-element enum types to boolean parameters, unless the meaning of the boolean is clear from the method name. Enums make your code easier to read and to write.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/LaunchMode.java
```diff
@@ -0,0 +1,17 @@
+/*******************************************************************************
+* Copyright (c) 2017 Microsoft Corporation and others.
```
yaohaizh: 2017? Really?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
```diff
@@ -113,6 +116,19 @@ private void initialize() {
     }
 
     private void registerHandler(IDebugRequestHandler handler) {
+        registerHandler(handler, true, false);
+    }
+
+    private void registerHandler(IDebugRequestHandler handler, boolean forDebug, boolean forNoDebug) {
```
yaohaizh: Two boolean value is tedious for the API level.  Either one boolean value or enum value.  

Keep the function/method declaration succinct.

PR #205, opened 2018-08-15
Validate mainClass and projectName configs

testforstephen:

The requirement is from issue https://github.com/Microsoft/vscode-java-debug/issues/355


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainClassHandler.java
```diff
@@ -146,5 +248,53 @@ public boolean equals(Object o) {
...
+
+            return (this.projectName + "|" + this.mainClass).compareTo(o.projectName + "|" + o.mainClass);
+        }
+
+        private boolean isDefaultProject(String projectName) {
+            return StringUtils.isEmpty(projectName);
+        }
+    }
+
+    class ProposalItemComparator implements Comparator<ResolutionItem> {
```
andxu: use CompareToBuilder

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainClassHandler.java
```diff
@@ -40,19 +44,35 @@
 
 public class ResolveMainClassHandler {
     private static final Logger logger = Logger.getLogger(Configuration.LOGGER_NAME);
+    private static final String CLASSNAME_REGX = "([$\\w]+\\.)*[$\\w]+";
```
andxu: I think a better way is to validate the main Class in js side.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaDebugDelegateCommandHandler.java
```diff
@@ -50,6 +52,8 @@ public Object executeCommand(String commandId, List<Object> arguments, IProgress
             return UsageDataStore.getInstance().fetchAll();
         } else if (UPDATE_DEBUG_SETTINGS.equals(commandId)) {
             return DebugSettingUtils.configDebugSettings(arguments);
+        } else if (VALIDATE_LAUNCHCONFIG.equals(commandId)) {
+            return new ResolveMainClassHandler().validateLaunchConfig(arguments);
```
andxu: How about make class ResolveMainClassHandler a utility class(with a changed name) and not new ResolveMainClassHandler every time

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainClassHandler.java
testforstephen: I don't want to separate the validation logic to two parts, that's hard for reading. JS side is responsible for UI related logic, and JAVA side is for computing the result.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainClassHandler.java
```diff
@@ -114,10 +134,93 @@ public void acceptSearchMatch(SearchMatch match) {
...
+        }
+
+        return new ValidationResult(true);
+    }
+
+    private List<ResolutionItem> computeProposals(List<Object> arguments, final String mainClass, final String projectName) {
+        List<ResolutionItem> proposals = resolveMainClassCore(arguments);
+
+        Collections.sort(proposals, new ProposalItemComparator((ResolutionItem item) -> {
+            if (item.mainClass.equals(mainClass)) {
```
andxu: Objects.equals

testforstephen: Nice suggestions.

PR #207, opened 2018-08-16
Spawn a process directly for noDebug mode.

Eskibear:
For noDebug mode, 
* spawn a process directly without appending vmargs.
* not to send `initialized` event back to the front end to avoid receiving future `SetBreakingpoints` requests.

See:
https://github.com/Microsoft/vscode-java-debug/issues/351#issuecomment-412730985


Eskibear:
The above commit doesn't support to Run in Integrated/External Terminal. I'll give a simple bu not-so-graceful solution in the next commit.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/LaunchRequestHandler.java
```diff
@@ -166,7 +176,7 @@
             ListeningConnector listenConnector = connectors.get(0);
```
Eskibear: listenConnector is not needed for NO_DEBUG mode. Actually I think it would be better to extract the common part instead of using lots of ifs

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/LaunchRequestHandler.java
```diff
@@ -268,20 +284,45 @@
...
-            logger.info("Launching debuggee VM succeeded.");
+                logger.info("Launching debuggee VM succeeded.");
+            } else {
+                String[] cmds = constructLaunchCommands(launchArguments, false, null);
+                File workingDir = null;
+                if (launchArguments.cwd != null && Files.isDirectory(Paths.get(launchArguments.cwd))) {
+                    workingDir = new File(launchArguments.cwd);
+                }
+                Process process = Runtime.getRuntime().exec(cmds, envVars, workingDir);
+                new Thread() {
```
Eskibear: might better to wrap these logic into a PromiseLike function.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/LaunchRequestHandler.java
```diff
@@ -318,7 +359,9 @@ private static String parseMainClassWithoutModuleName(String mainClass) {
 
         List<String> launchCmds = new ArrayList<>();
         launchCmds.add(System.getProperty("java.home") + slash + "bin" + slash + "java");
-        launchCmds.add(String.format("-agentlib:jdwp=transport=dt_socket,server=%s,suspend=y,address=%s", serverMode ? "y" : "n", address));
+        if (StringUtils.isNotEmpty(address)) {
```
Eskibear: `private String[] constructLaunchCommands(LaunchArguments launchArguments, boolean serverMode, String address)` can be improved. Because when `address` is null, the value of `serverMode` will not make any difference.  Will it be better to use something like private `String[] constructLaunchCommands(LaunchArguments launchArguments, String debugArgs)`

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/NoDebugSession.java
```diff
@@ -0,0 +1,79 @@
...
+*******************************************************************************/
+
+package com.microsoft.java.debug.core;
+
+import java.util.List;
+
+import com.sun.jdi.ThreadReference;
+import com.sun.jdi.VirtualMachine;
+
+public class NoDebugSession implements IDebugSession {
```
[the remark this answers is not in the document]
testforstephen: DummyDebugSession

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/LaunchRequestHandler.java
```diff
@@ -166,7 +176,7 @@
             ListeningConnector listenConnector = connectors.get(0);
             Map<String, Connector.Argument> args = listenConnector.defaultArguments();
             ((Connector.IntegerArgument) args.get("timeout")).setValue(ACCEPT_TIMEOUT);
-            String address = listenConnector.startListening(args);
+            String address = context.getLaunchMode() == LaunchMode.DEBUG ? listenConnector.startListening(args) : null;
 
             String[] cmds = constructLaunchCommands(launchArguments, false, address);
             RunInTerminalRequestArguments requestArgs = null;
```
testforstephen: Extract the logic about executing the launchInTerminal request to a helper method.

`CompletableFuture<void> sendLaunchInTerminalRequest(LaunchArguments launchArguments, IDebugAdapterContext context)`

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/LaunchRequestHandler.java
```diff
@@ -166,7 +176,7 @@
             ListeningConnector listenConnector = connectors.get(0);
             Map<String, Connector.Argument> args = listenConnector.defaultArguments();
             ((Connector.IntegerArgument) args.get("timeout")).setValue(ACCEPT_TIMEOUT);
-            String address = listenConnector.startListening(args);
+            String address = context.getLaunchMode() == LaunchMode.DEBUG ? listenConnector.startListening(args) : null;
```
andxu: add a method isDebugMode in context

Eskibear: indeed

Eskibear: Revert to avoid confusion caused by the name "DebugMode"

Eskibear:
Per discussion with @testforstephen , "Start Without Debugging" is a much more simple than "Start Debugging", it should only respond to `initilaize` (same logic), `launch`, `disconnect` requests. It's better not to mess up the handlers for "debug mode". And based on the fact that we can only distinguish debug/noDebug in `launch` request, following steps for the 2 launch modes will be:
* Share the same InitializeRequestHandler.
* Use 2 different LaunchRequestHandler for debug/noDebug mode respectively, the common code can be put in a super class.
* Keep other RequestHandlers for debug mode as before.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AbstractLaunchRequestHandler.java
```diff
@@ -0,0 +1,241 @@
...
+            );
+        }
+
+        return resultFuture;
+    }
+
+    protected abstract Process launchInternalDebuggeeProcess(LaunchArguments launchArguments, IDebugAdapterContext context)
+            throws IOException, IllegalConnectorArgumentsException, VMStartException;
+
+    protected String[] constructEnvironmentVaraiables(LaunchArguments launchArguments) {
```
testforstephen: typos. constructEnvironmentVar **a** iables

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StartWithoutDebuggingRequestsHandler.java
```diff
@@ -0,0 +1,168 @@
...
+
+    @Override
+    protected CompletableFuture<Response> launchInTerminal(LaunchArguments launchArguments, Response response,
+            IDebugAdapterContext context) {
+        CompletableFuture<Response> resultFuture = new CompletableFuture<>();
+
+        final String launchInTerminalErrorFormat = "Failed to launch debuggee in terminal. Reason: %s";
+
+        String[] cmds = constructLaunchCommands(launchArguments, false, null);
+        RunInTerminalRequestArguments requestArgs = null;
```
testforstephen: The logic to construct launchInTerminal request arguments and handle the response is similar as the normal LaunchRequestHandler, suggest to extract it to a helper method in the superclass.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StartWithoutDebuggingRequestsHandler.java
```diff
@@ -0,0 +1,168 @@
...
+import com.sun.jdi.connect.IllegalConnectorArgumentsException;
+import com.sun.jdi.connect.VMStartException;
+
+public class StartWithoutDebuggingRequestsHandler extends AbstractLaunchRequestHandler {
+    private static final String TERMINAL_TITLE = "Java Process Console";
+    private Process debuggeeProcess;
+
+    @Override
+    public List<Command> getTargetCommands() {
+        return Arrays.asList(Command.values());
```
testforstephen: Add a comment that all requests will be dispatched to this handler.
Or extract `Command.values()` to a private method `getAllDebugCommands()` for self explain.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StartWithoutDebuggingRequestsHandler.java
```diff
@@ -0,0 +1,168 @@
...
+import com.microsoft.java.debug.core.protocol.Requests.Arguments;
+import com.microsoft.java.debug.core.protocol.Requests.CONSOLE;
+import com.microsoft.java.debug.core.protocol.Requests.Command;
+import com.microsoft.java.debug.core.protocol.Requests.DisconnectArguments;
+import com.microsoft.java.debug.core.protocol.Requests.LaunchArguments;
+import com.microsoft.java.debug.core.protocol.Requests.RunInTerminalRequestArguments;
+import com.sun.jdi.connect.IllegalConnectorArgumentsException;
+import com.sun.jdi.connect.VMStartException;
+
+public class StartWithoutDebuggingRequestsHandler extends AbstractLaunchRequestHandler {
```
testforstephen: also explicitly declare it `implements IDebugRequestHandler`

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StartWithoutDebuggingRequestsHandler.java
```diff
@@ -0,0 +1,168 @@
...
+            return launchArguments.noDebug ? handleLaunchCommand(arguments, response, context) : CompletableFuture.completedFuture(response);
+        } else if (Command.DISCONNECT.equals(command)) {
+            return handleDisconnectCommand(arguments, response, context);
+        } else {
+            if (context.getLaunchMode() == LaunchMode.NO_DEBUG) {
+                String errorMsg = String.format("Unrecognized request: { _request: %s }", command);
+                throw AdapterUtils.createCompletionException(errorMsg, ErrorCode.UNRECOGNIZED_REQUEST_FAILURE);
+            }
+            return CompletableFuture.completedFuture(response);
+        }
```
testforstephen: Add a method to detect if the launchMode is noDebug or not.
```
private boolean isNoDebug(Command command, Arguments arguments, IDebugAdapterContext context) {
    return context.getLaunchMode() == LaunchMode.NO_DEBUG || (Command.LAUNCH == command && ((LaunchArgument) arguments).noDebug);
}
```

Then the `handle` logic could be 
```
if (isNoDebug()) {
    if (launch) ...
    else if (disconnect) ...
    else unrecognized request...
}
do nothing and return response directly...
```

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StartWithoutDebuggingRequestsHandler.java
Eskibear: Agree. Suggest to do that in another new PR.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AbstractLaunchRequestHandler.java
```diff
@@ -0,0 +1,241 @@
...
+import com.sun.jdi.connect.IllegalConnectorArgumentsException;
+import com.sun.jdi.connect.VMStartException;
+
+public abstract class AbstractLaunchRequestHandler implements IDebugRequestHandler {
+    protected static final Logger logger = Logger.getLogger(Configuration.LOGGER_NAME);
+    protected static final long RUNINTERMINAL_TIMEOUT = 10 * 1000;
+
+    @Override
+    public List<Command> getTargetCommands() {
+        return Arrays.asList(Command.LAUNCH);
```
yaohaizh: This should be abstract method since this adapter doesn't handle any specific command.

Eskibear: Good catch. And also, since it actually doesn't implements `IDebugRequestHandler`, do you think we should remove the declaration here, but add in derived classes?

Eskibear:
Up to fe08cef, the design is as following:

image

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/CommonLaunchRequestHandler.java
```diff
@@ -0,0 +1,197 @@
...
+import com.microsoft.java.debug.core.protocol.Messages.Response;
+import com.microsoft.java.debug.core.protocol.Requests.Arguments;
+import com.microsoft.java.debug.core.protocol.Requests.CONSOLE;
+import com.microsoft.java.debug.core.protocol.Requests.Command;
+import com.microsoft.java.debug.core.protocol.Requests.LaunchArguments;
+
+public class CommonLaunchRequestHandler implements IDebugRequestHandler {
+    protected static final Logger logger = Logger.getLogger(Configuration.LOGGER_NAME);
+    protected static final long RUNINTERMINAL_TIMEOUT = 10 * 1000;
+    private ILaunchDelegateHandler launchWithDebuggingHandler = new LaunchRequestHandler();
```
yaohaizh: One private field is enough in this case? Delay the construction time

Eskibear: Is it this way?
```
private ILaunchDelegateHandler activeLaunchHandler;

activeLaunchHandler = launchArguments.noDebug 
? new LaunchWithoutDebuggingRequestHandler() 
: new LaunchRequestHandler();
```

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/LaunchWithoutDebuggingRequestHandler.java
```diff
@@ -0,0 +1,169 @@
...
+import com.sun.jdi.connect.VMStartException;
+
+public class LaunchWithoutDebuggingRequestHandler implements ILaunchDelegateHandler {
+    protected static final Logger logger = Logger.getLogger(Configuration.LOGGER_NAME);
+    protected static final String TERMINAL_TITLE = "Java Process Console";
+    protected static final long RUNINTERMINAL_TIMEOUT = 10 * 1000;
+
+    private Process launchInternalDebuggeeProcess(LaunchArguments launchArguments, IDebugAdapterContext context)
+            throws IOException, IllegalConnectorArgumentsException, VMStartException {
+        String[] cmds = CommonLaunchRequestHandler.constructLaunchCommands(launchArguments, false, null);
```
andxu: why not make CommonLaunchRequestHandler  abstract class: AbstractLaunchRequestHandler  and extends it?

Eskibear: Previous way was using the abstract class you mentioned, and there would be 2 derived LaunchRequestHandler for debug/noDebug mode respectively. If so, we have to register both handlers to respond to `launch` requests. (because we never know the launching mode before we handle the request)
By following delegation pattern, `launch` request is handled by the CommonLaunchRequestHandler, according to the noDebug argument, it then calls interfaces in different delegates. By putting those abstract methods to an interface, it's also clearer and easier to read.

Eskibear:
@andxu Please take a look. After renaming: 
image

PR #211, opened 2018-08-28
record sum of responding time for the debug session

andxu:


andxu:
We should record the duration in ms for each command to have the ability to know the user's sufferings like https://github.com/Microsoft/vscode-java-debug/issues/305

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/UsageDataSession.java
```diff
@@ -139,6 +141,7 @@ public void submitUsageData() {
         props.put("commandCount", JsonUtils.toJson(commandCountMap));
         props.put("breakpointCount", JsonUtils.toJson(breakpointCountMap));
         props.put("userErrorCount", JsonUtils.toJson(userErrorCount));
+        props.put("commandPerCount", JsonUtils.toJson(commandPerfCountMap));
```
Eskibear: commandPerCount vs commandPerfCountMap, missing "f"

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/UsageDataSession.java
```diff
@@ -114,6 +115,7 @@ public void recordResponse(Response response) {
                 requestEventMap.remove(response.request_seq);
             }
             long duration = responseMillis - requestMillis;
+            commandPerfCountMap.compute(command, (k, v) -> (v == null ? 0 : v.intValue()) + (int) duration);
```
Eskibear: It stores total duration of a certain command. Will average duration be more useful?

PR #219, opened 2018-09-06
Fix issue #402 missing runtime scope classpath entry

andxu:
Redhat.java update to version 0.30.0 broke launching/debugging of Spring boot-apps 
#402


andxu:
https://github.com/Microsoft/vscode-java-debug/issues/402

PR #222, opened 2018-09-06
Add a utility to resolve main methods from a java file

testforstephen:

This is for the issue https://github.com/Microsoft/vscode-java-debug/issues/375

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
@@ -1,0 +1,139 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Microsoft Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Microsoft Corporation - initial API and implementation
+ *******************************************************************************/
+
+package com.microsoft.java.debug.plugin.internal;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.jdt.core.Flags;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaElement;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IMethod;
+import org.eclipse.jdt.core.ISourceRange;
+import org.eclipse.jdt.core.ISourceReference;
+import org.eclipse.jdt.core.IType;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.ls.core.internal.JDTUtils;
+import org.eclipse.jdt.ls.core.internal.managers.ProjectsManager;
+import org.eclipse.lsp4j.Range;
+
+public class ResolveMainMethodHandler {
+    /**
+     * Resolve the main methods from the current file.
+     * @return an array of main methods.
+     */
+    public static Object resolveMainMethods(List<Object> arguments) throws Exception {
+        if (arguments == null || arguments.size() == 0) {
+            return Collections.emptyList();
+        }
+
+        String uri = (String) arguments.get(0);
+        final ICompilationUnit unit = JDTUtils.resolveCompilationUnit(uri);
+        if (unit == null || unit.getResource() == null || !unit.getResource().exists()) {
+            return Collections.emptyList();
+        }
+
+        try {
+            return resolveMainMethodCore(unit);
+        } catch (Exception e) {
+            throw new Exception("Failed to resolve main method codeLens: " + e.getMessage(), e);
+        }
+    }
+
+    private static List<MainMethod> resolveMainMethodCore(ICompilationUnit typeRoot) throws JavaModelException {
+        List<MainMethod> mainMethods = new ArrayList<>();
+
+        List<IMethod> jdtMainMethods = new ArrayList<>();
+        IType[] topLevelTypes = typeRoot.getTypes();
+        for (IType topType : topLevelTypes) {
+            jdtMainMethods.addAll(resolveMainMethodCore(topType, 1));
+        }
+
+        for (IMethod method : jdtMainMethods) {
+            MainMethod mainMethod = constructMainMethod(typeRoot, method);
+            if (mainMethod != null) {
+                mainMethods.add(mainMethod);
+            }
+        }
+
+        return mainMethods;
+    }
+
+    private static List<IMethod> resolveMainMethodCore(IType type, int level) throws JavaModelException {
+        // main method can only exist in the static class or top level class.
+        if (level > 1 && !Flags.isStatic(type.getFlags())) {
+            return Collections.emptyList();
+        }
+
+        List<IMethod> mainMethods = new ArrayList<>();
+
+        // Have at most one main method at the same level.
+        for (IMethod method : type.getMethods()) {
+            if (method.isMainMethod()) {
+                mainMethods.add(method);
+                break;
+            }
+        }
+
+        for (IType child : type.getTypes()) {
+            mainMethods.addAll(resolveMainMethodCore(child, level + 1));
+        }
+
+        return mainMethods;
+    }
+
+    private static Range getRange(ICompilationUnit typeRoot, IJavaElement element) throws JavaModelException {
+        ISourceRange r = ((ISourceReference) element).getNameRange();
+        final Range range = JDTUtils.toRange(typeRoot, r.getOffset(), r.getLength());
+        return range;
+    }
+
+    private static MainMethod constructMainMethod(ICompilationUnit typeRoot, IMethod method) throws JavaModelException {
+        final Range range = getRange(typeRoot, method);
+        IResource resource = typeRoot.getResource();
+        if (resource != null) {
+            IProject project = resource.getProject();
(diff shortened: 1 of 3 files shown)


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
```diff
@@ -0,0 +1,139 @@
...
+import org.eclipse.jdt.ls.core.internal.JDTUtils;
+import org.eclipse.jdt.ls.core.internal.managers.ProjectsManager;
+import org.eclipse.lsp4j.Range;
+
+public class ResolveMainMethodHandler {
+    /**
+     * Resolve the main methods from the current file.
+     * @return an array of main methods.
+     */
+    public static Object resolveMainMethods(List<Object> arguments) throws Exception {
```
andxu: DebugException?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
```diff
@@ -0,0 +1,139 @@
...
+        try {
+            return resolveMainMethodCore(unit);
+        } catch (Exception e) {
+            throw new Exception("Failed to resolve main method codeLens: " + e.getMessage(), e);
+        }
+    }
+
+    private static List<MainMethod> resolveMainMethodCore(ICompilationUnit typeRoot) throws JavaModelException {
+        List<MainMethod> mainMethods = new ArrayList<>();
+
```
andxu: a line break is not needed

andxu: The logic is not clear to understand, eclipse has some examples: org.eclipse.jdt.internal.launching.JavaLaunchableTester, you can first get the location and then lazy check the runtime configurations until the on the click.

```
/**
	 * gets the type of the IJavaElement
	 * @param element the element to inspect
	 * @return the type
	 */
	private IType getType(IJavaElement element) {
        IType type = null;
        if (element instanceof ICompilationUnit) {
            type= ((ICompilationUnit) element).findPrimaryType();
        }
		else if (element instanceof IOrdinaryClassFile) {
			type = ((IOrdinaryClassFile) element).getType();
        }
        else if (element instanceof IType) {
            type = (IType) element;
        }
        else if (element instanceof IMember) {
            type = ((IMember)element).getDeclaringType();
        }
        return type;
    }

	/**
	 * Determines is the java element contains a main method.
	 *
	 * @param element the element to check for the method
	 * @return true if the method is found in the element, false otherwise
	 */
	private boolean hasMain(IJavaElement element) {
		try {
            IType type = getType(element);
            if(type != null && type.exists()) {
				if(hasMainMethod(type)) {
					return true;
				}
				//failed to find in public type, check static inner types
				IJavaElement[] children = type.getChildren();
				for(int i = 0; i < children.length; i++) {
					if(hasMainInChildren(getType(children[i]))) {
						return true;
					}
				}
            }
		}
		catch (JavaModelException e) {}
		catch (CoreException ce){}
		return false;
	}

	/**
	 * Returns if the specified <code>IType</code> has a main method
	 * @param type the type to inspect for a main type
	 * @return true if the specified type has a main method, false otherwise
	 * @throws JavaModelException if there is an error in the backing Java model
	 * @since 3.3
	 */
	private boolean hasMainMethod(IType type) throws JavaModelException {
		IMethod[] methods = type.getMethods();
		for (int i= 0; i < methods.length; i++) {
			if(methods[i].isMainMethod()) {
				return true;
			}
		}
		return false;
	}

	/**
	 * This method asks the specified <code>IType</code> if it has a main method, if not it recurses through all of its children
	 * When recursing we only care about child <code>IType</code>s that are static.
	 * @param type the <code>IType</code> to inspect for a main method
	 * @return true if a main method was found in specified <code>IType</code>, false otherwise
	 * @throws CoreException if there is an error
	 * @since 3.3
	 */
	private boolean hasMainInChildren(IType type) throws CoreException {
		if(type.isClass() & Flags.isStatic(type.getFlags())) {
			if(hasMainMethod(type)) {
				return true;
			}
			IJavaElement[] children = type.getChildren();
			for(int i = 0; i < children.length; i++) {
				if(children[i].getElementType() == IJavaElement.TYPE) {
					return hasMainInChildren((IType) children[i]);
				}
			}
		}
		return false;
	}
```

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
testforstephen: fixed.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
testforstephen: Refined the main method searching logic. `MainMethodSearchEngine` is suitable for general searching purpose. For the specific java file, i prefer to iterate the member list to find the main method.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
```diff
@@ -0,0 +1,161 @@
...
+                return new MainMethod(range, mainClass, projectName);
+            }
+        }
+
+        return null;
+    }
+
+    private static Range getRange(ICompilationUnit typeRoot, IJavaElement element) throws JavaModelException {
+        ISourceRange r = ((ISourceReference) element).getNameRange();
+        final Range range = JDTUtils.toRange(typeRoot, r.getOffset(), r.getLength());
```
andxu: return  JDTUtils.toRange(typeRoot, r.getOffset(), r.getLength());

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
```diff
@@ -0,0 +1,161 @@
...
+
+import com.microsoft.java.debug.core.DebugException;
+
+public class ResolveMainMethodHandler {
+    /**
+     * Resolve the main methods from the current file.
+     * @return an array of main methods.
+     */
+    public static Object resolveMainMethods(List<Object> arguments) throws DebugException {
+        if (arguments == null || arguments.size() == 0) {
```
andxu: isEmpty

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
```diff
@@ -0,0 +1,161 @@
...
+
+        try {
+            return resolveMainMethodCore(unit);
+        } catch (Exception e) {
+            throw new DebugException("Failed to resolve main method codeLens: " + e.getMessage(), e);
+        }
+    }
+
+    private static List<MainMethod> resolveMainMethodCore(ICompilationUnit compilationUnit) throws JavaModelException {
+        List<MainMethod> codeLenses = new ArrayList<>();
```
andxu: confuse name of codeLenses  codeLens

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
```diff
@@ -0,0 +1,161 @@
...
+        List<IType> potentialTypes = getPotentialMainClassTypes(compilationUnit.getTypes(), 1);
+        List<IMethod> result = new ArrayList<>();
+        for (IType type: potentialTypes) {
+            result.addAll(searchMainMethodsInType(type));
+        }
+
+        return result;
+    }
+
+    private static List<IMethod> searchMainMethodsInType(IType type) throws JavaModelException {
```
andxu: At most one in the list, change it to IMethod

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainMethodHandler.java
```diff
@@ -0,0 +1,162 @@
...
+
+        return null;
+    }
+
+    private static Range getRange(ICompilationUnit typeRoot, IJavaElement element) throws JavaModelException {
+        ISourceRange r = ((ISourceReference) element).getNameRange();
+        return JDTUtils.toRange(typeRoot, r.getOffset(), r.getLength());
+    }
+
+    static class MainMethod {
```
yaohaizh: Better utilize the org.eclipse.lsp4j if possible to make the design consistently.

testforstephen: I saw there is a `CodeLens` type in lsp4j, it's suitable for Java language server, because all codelens operation (label, command) are implemented in Java code. But in Java debugger, we just take the Java code as an utility method, the action command is implemented in ts. If we put the codeLens command info in Java code, it looks like not clean.

yaohaizh: Then we should avoid use CodeLens in the java code for debugging, just maininfo or whatever. Otherwise, this will  mix concept in the code.

testforstephen: Agree. Have changed the codeLens to mainMethod info.

PR #223, opened 2018-09-07
Remove duplicate classpath

andxu:


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveClasspathsHandler.java
```diff
@@ -189,7 +191,10 @@ public void acceptSearchMatch(SearchMatch match) {
                     }
                     String location = entries[j].getLocation();
                     if (location != null) {
-                        resolved.add(location);
+                        // remove duplicate classpath
+                        if (!resolved.contains(location)) {
```
yaohaizh: The contains is not necessary.  https://docs.oracle.com/javase/7/docs/api/java/util/HashSet.html#add(E)

yaohaizh: Should we normalize the path?

Issue #134, opened 2017-12-18
"continue" response sent after the following "stopped" event - should be opposite


offer8:
1. Put 2 breakpoints (e.g lines: 28, 38) in Java method
2. Stopped a the first breakpoint - OK as expected  
3. Sent "resume" command
4. Sometimes the "continue" response sent after the following "stopped" event - so our debugger client doesn't stopped at second breakpoint.

Following is the log:
2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR {"command":"continue","type":"request","seq":"17","arguments":{"threadId":13}}

2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR 18/12/2017 11:38:53.398 - [com.microsoft.java.debug.core.EventHub.lambda$start$28] - [FINE] - 

2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR JDI Event: BreakpointEvent@com.company.mta123.simple.HelloServlet:38 in thread http-nio-0.0.0.0-3000-exec-3
2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR 18/12/2017 11:38:53.399 - [com.microsoft.java.debug.
2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR 2017/12/18 11:38:53  core.protocol.AbstractProtocolServer.sendMessage] - [FINE] - 
2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR [[EVENT]]
2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR Content-Length: 131
**2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR {"event":"stopped","body":{"threadId":13,"reason":"breakpoint","allThreadsStopped":false,"type":"stopped"},"seq":25,"type":"event"}**

2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR 18/12/2017 11:38:53.403 - [com.microsoft.java.debug.core.protocol.AbstractProtocolServer.sendMessage] - [FINE] - 
2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR [[RESPONSE]]
2017-12-18T13:38:53.48+0200 [APP/PROC/WEB/0]ERR Content-Length: 118
[...]

PR #227, opened 2018-09-19
Fix debug buttons disabled issue: send StoppedEvent after StepResponse/ContinueResponse

testforstephen:

It fixes the bug https://github.com/Microsoft/vscode-java-debug/issues/411 and https://github.com/Microsoft/java-debug/issues/134

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/ProtocolServer.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/ProtocolServer.java
@@ -15,6 +15,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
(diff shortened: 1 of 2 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/ProtocolServer.java
```diff
@@ -75,43 +80,83 @@ public void sendResponse(Messages.Response response) {
...
+                    }
+
+                    sendResponse(AdapterUtils.setErrorResponse(response,
+                            errorCode,
+                            exceptionMessage));
+                }
+                return null;
+            }).join();
+        } finally {
+            synchronized (this) {
```
yaohaizh: Avoid lock the whole object, use private object for lock.

testforstephen: fixed.

PR #237, opened 2018-10-08
fix #440: The startup command arguments for internalConsole are different from integratedTerminal

andxu:
https://github.com/Microsoft/vscode-java-debug/issues/440


andxu:
The parseArguments jinbo wrote is different from the DebugPlugin.parseArguments which will cause the startup command line argument is different when using different console options:
integratedTerminal and internalConsole

PR #253, opened 2019-02-14
Provide jarmanifest/argfile approachs to shorten the command line

testforstephen:

Fix the bug https://github.com/Microsoft/vscode-java-debug/issues/110.

Provide two kinds of approaches jarmanifest and argfile to shorten the command line length.


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Requests.java
```diff
@@ -60,6 +60,12 @@
         externalTerminal;
     }
 
+    public static enum CliStyle {
```
Eskibear: > Because they are constants, the names of an enum type's fields are in uppercase letters.

See https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html for reference

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Requests.java
```diff
@@ -72,6 +78,7 @@
         public boolean stopOnEntry;
         public boolean noDebug = false;
         public CONSOLE console = CONSOLE.internalConsole;
+        public CliStyle shortenCommandLine = CliStyle.none;
```
Eskibear: better naming for `shortenCommandLine` , maybe simply `cliStyle`. 

Because I see statement ` if (launchArguments.shortenCommandLine == CliStyle.argfile)` in other files, it confuses me that a `shortenCommandLine` equals to a kind of `CliStyle`

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
```diff
@@ -72,8 +72,7 @@ public DebugAdapter(IProtocolServer server, IProviderContext providerContext) {
         Command command = Command.parse(request.command);
         Arguments cmdArgs = JsonUtils.fromJson(request.arguments, command.getArgumentType());
 
-        if (debugContext.isVmTerminated()) {
-            // the operation is meaningless
+        if (debugContext.isVmTerminated() && command != Command.DISCONNECT) {
```
Eskibear: curious about what does `command != Command.DISCONNECT` fix?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Requests.java
testforstephen: Nice.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
testforstephen: Don't intercept disconnect request here because we want to leave it to disconnectHandler to cleanup temp launch files.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AbstractDisconnectRequestHandler.java
```diff
@@ -0,0 +1,97 @@
...
+                    Files.deleteIfExists(context.getArgsfile());
+                    context.setArgsfile(null);
+                }
+            } catch (IOException e) {
+                // do nothing.
+                logger.log(Level.WARNING, "Failed to destory launch files, will retry again.");
+            }
+
+            try {
+                Thread.sleep(1000);
```
akaroml: TimeUnit.SECONDS.sleep(1);

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/LaunchRequestHandler.java
```diff
@@ -87,6 +97,62 @@
 
         activeLaunchHandler.preLaunch(launchArguments, context);
 
+        Path tempfile = null;
```
akaroml: The body of this function is already long. Please refactor by putting the procedures of generating Jar Manifest and ArgFile into different utility functions. Those functions should be functional without referencing states from other parts. After that the body looks like:

```java
if (launchArguments.shortenCommandLine == ShortenApproach.JARMANIFEST) {
    Path path = AdapterUtils.generateClasspathJar(...);
    context.setClasspathJar(path);
} else if (...) {
}
```

PR #256, opened 2019-02-28
Increase the timeout when attaching to integrated/external terminal

testforstephen:

Fixes Microsoft/vscode-java-debug#541


yaohaizh:
Could we listening to event for the TIMEOUT value. I don't think it is reasonable to have hard code value for this case. If user double the path content, the 20 seconds will not work, right?

testforstephen:
@yaohaizh what event do you want to listening to for computing the timeout value?

testforstephen:
Refine the timeout error message if the command line is a little long.

PR #261, opened 2019-03-21
Add completion support for classfile.

yaohaizh:
Signed-off-by: Yaohai Zheng <yaozheng@microsoft.com>


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/CompletionProposalRequestor.java
```diff
@@ -0,0 +1,257 @@
...
+import org.eclipse.jdt.ls.core.internal.handlers.CompletionResponses;
+import org.eclipse.lsp4j.CompletionItem;
+import org.eclipse.lsp4j.CompletionItemKind;
+
+import com.google.common.collect.ImmutableSet;
+import com.microsoft.java.debug.core.Configuration;
+
+
+@SuppressWarnings("restriction")
+public final class CompletionProposalRequestor extends CompletionRequestor {
```
testforstephen: What's the difference with the origin copy?

PR #264, opened 2019-04-12
Add utility to check java project settings

testforstephen:

This is the utility for the PR https://github.com/Microsoft/vscode-java-debug/pull/561

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaDebugDelegateCommandHandler.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaDebugDelegateCommandHandler.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2017 Microsoft Corporation and others.
+* Copyright (c) 2017-2019 Microsoft Corporation and others.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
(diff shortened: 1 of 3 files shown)


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaDebugDelegateCommandHandler.java
```diff
@@ -30,6 +30,7 @@
     public static final String VALIDATE_LAUNCHCONFIG = "vscode.java.validateLaunchConfig";
     public static final String RESOLVE_MAINMETHOD = "vscode.java.resolveMainMethod";
     public static final String INFER_LAUNCH_COMMAND_LENGTH = "vscode.java.inferLaunchCommandLength";
+    public static final String DETECT_PREVIEW_FLAG = "vscode.java.detectEnablePreviewFeaturesFlag";
```
yaohaizh: I think this could be a more common API about fetch settings.

testforstephen: The new API will return the java project settings map, and delay the check to the client side.

yaohaizh: This line is not necessary now.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ProjectConfigHandler.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.jdt.core.IJavaProject;
+
+public class ProjectConfigHandler {
+    /**
+     * Return the java project settings.
+     * @param arguments - The first argument is the project name, the second argument indicate whether to get the inherited options.
+     * @return The map of current settings of all options(key type: String; value type: String)
+     */
```
yaohaizh: I think API should be flexible for multiple purpose yet concise. 

In here, every variable has its own API, which will bloat API number. We should keep the API design simple.

testforstephen: addressed. Will change the api to check project settings. The api parameter is `className: String, projectName: String, inheritedOptions: boolean, expectedOptions: Map<String, String>`

PR #268, opened 2019-04-16
Fix variable perf issue & logical structure doesn't work with JDK 11

testforstephen:

fixes Microsoft/vscode-java-debug#571
fixes Microsoft/vscode-java-debug#568

Use the evaluation engine to resolve the logical structure, it has perf issue. On the other hand, in JDK 11, jdt has an issue on parsing the compiled expression `return entrySet();` for Map. See jdt issue https://bugs.eclipse.org/bugs/show_bug.cgi?id=543604.

The PR will get the logical structure via the JDI approach. It will fix perf issue and JDK 11 problem.

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/VariablesRequestHandler.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/VariablesRequestHandler.java
@@ -37,8 +37,8 @@ import com.microsoft.java.debug.core.adapter.IEvaluationProvider;
 import com.microsoft.java.debug.core.adapter.IStackFrameManager;
 import com.microsoft.java.debug.core.adapter.variables.IVariableFormatter;
 import com.microsoft.java.debug.core.adapter.variables.JavaLogicalStructure;
-import com.microsoft.java.debug.core.adapter.variables.JavaLogicalStructure.Expression;
-import com.microsoft.java.debug.core.adapter.variables.JavaLogicalStructure.ExpressionType;
+import com.microsoft.java.debug.core.adapter.variables.JavaLogicalStructure.LogicalStructureExpression;
+import com.microsoft.java.debug.core.adapter.variables.JavaLogicalStructure.LogicalStructureExpressionType;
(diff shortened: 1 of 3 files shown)


yaohaizh:
Please split to two PRs for each fix.

testforstephen:
The same root cause for this two problems, so one fix is ok.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/variables/JavaLogicalStructure.java
```diff
@@ -98,8 +96,26 @@ public String getName() {
             return name;
         }
 
-        public String getValue() {
+        public Expression getValue() {
             return value;
         }
     }
+
+    public static class Expression {
```
andxu: Don't use Expression as the class name since it will be confusing with JDT classes, use a

andxu: inner class of JavaLogicalStructure

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/VariablesRequestHandler.java
```diff
@@ -244,6 +255,53 @@
...
+        if (expression.type == LogicalStructureExpressionType.METHOD) {
+            return getValueByInvokeMethod(thisObject, expression.value, thread);
+        } else if (expression.type == LogicalStructureExpressionType.FIELD) {
+            return getValueByField(thisObject, expression.value, thread);
+        } else {
+            return evaluationEngine.evaluate(expression.value, thisObject, thread).get();
+        }
+    }
+
+    private Value getValueByInvokeMethod(ObjectReference thisObject, String methodName, ThreadReference thread)
```
andxu: getValueByInvokeMethod => getValueByMethod

andxu: Move this method to LogicalStructureExpression

Issue #279, opened 2019-06-12
[performance] should break out the while loop


tom-shan:
https://github.com/microsoft/java-debug/blob/master/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AbstractDisconnectRequestHandler.java#L73-L94
Now it will costs about 5 seconds in the destroyLaunchFiles function, maybe there should have a break statement if no IOException catched in each loop for better performance.

testforstephen:
nice finding.

testforstephen:
@tom-shan Could you help make a pull request?

PR #280, opened 2019-06-12
Break out of the while loop for better performance

tom-shan:
Now it will cost about 5 seconds in the destroyLaunchFiles function,
there should have a break statement if no IOException is caught
in each loop iteration for better performance.

Signed-off-by: tom-shan <s00347147@163.com>


tom-shan:
fix https://github.com/microsoft/java-debug/issues/279
@testforstephen please have a review

testforstephen:
Looks good to me.
Thanks for the contribution.

PR #284, opened 2019-06-28
skip cleanup when attached to the process

marek1840:
Currently, the Debug Adapter, after attaching to the process and receiving a `disconnect` request, performs  a blocking cleanup operation. 

This operation blocks for 5 seconds to do a `launch` related cleanup, so it makes sense to skip it in the case of an attached Debug Adapter.

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AbstractDisconnectRequestHandler.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AbstractDisconnectRequestHandler.java
@@ -56,7 +56,7 @@ public abstract class AbstractDisconnectRequestHandler implements IDebugRequestH
     }
 
     private boolean shouldDestroyLaunchFiles(IDebugAdapterContext context) {
-        if(context.isAttached()){
+        if (context.isAttached()) {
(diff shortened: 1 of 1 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/AbstractDisconnectRequestHandler.java
```diff
@@ -56,6 +56,10 @@ private void destroyResource(IDebugAdapterContext context) {
     }
 
     private boolean shouldDestroyLaunchFiles(IDebugAdapterContext context) {
+        if(context.isAttached()){
```
testforstephen: a whitepace is needed between `if` and parenthesis `(`.
`if ()`

testforstephen:
thanks for contribution. an earlier PR https://github.com/microsoft/java-debug/pull/280 should cover this scenario well. 
if it's attach mode, classpathJar and argsFile don't exist, so it will break the retry loop directly. looks like this PR is not necessary.

marek1840:
> thanks for contribution. an earlier PR #280 should cover this scenario well.
> if it's attach mode, classpathJar and argsFile don't exist, so it will break the retry loop directly. looks like this PR is not necessary.

Great, thanks!

PR #285, opened 2019-07-02
Unblock run method

### LINKED ISSUE #286: Server thread blocks until the client disconnects
The `run` method in `AbstractProtocolServer` is blocking the thread on this line: [AbstractProtocolServer:92](https://github.com/microsoft/java-debug/blob/master/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/AbstractProtocolServer.java#L92)

In my case, I may have to close the Debug Server in the middle of debugging while the whole application (within which the Debug Server operates) is still running, so I would like to do so as gracefully as I can.

Currently, it can be done by closing the I/O streams, as nothing else will unblock the thread on which the server is running. This approach results in an exception being throws in the server thread (i.e. socket closed) which is not something I would expect when closing the server.

Of course, the exception can be catched and ignored but it introduces an unnecessary level of complexity (to recognize it is an "expected exception" and not an actual issue).

marek1840:
Currently there is no nice way to stop the running server.
This makes the thread on which the server is running interruptible. 

After the server is stopped and the currently processed request is handled, the debug adapter is being closed, which means:
1. The debug session is terminated/detached
2. The resources created when launching are removed

Please note, that 2. was achieved by moving the logic from `AbstractDisconnectRequestHandler` directly to the `DebugAdapterContext`. If that is not acceptable, please guide me how it should be done.

Closes #286


testforstephen:
@marek1840 thanks for the contribution. Could you open an issue to explain the scenario why this is needed?

marek1840:
Closing, since the workaround for #286 has been suggested

PR #287, opened 2019-08-05
Check style only on Travis CI

Eskibear:
Do not force checkstyle check locally, as some devs (Windows) checkout code as CRLF.


pom.xml
```diff
@@ -148,6 +139,19 @@
             </plugins>
         </pluginManagement>
     </build>
+    <profiles>
+		<profile>
```
jdneo: Use space

Eskibear:
@jdneo indentation updated.

PR #296, opened 2019-09-12
Add command to check if the hovered element is main method

testforstephen:


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveElementHandler.java
```diff
@@ -0,0 +1,73 @@
...
+import com.microsoft.java.debug.core.DebugException;
+
+public class ResolveElementHandler {
+
+    /**
+     * Resolve the Java element at the selected position.
+     * @return the resolved Java element information.
+     */
+    public static Object resolveElementAtSelection(List<Object> arguments, IProgressMonitor monitor) throws DebugException {
+        if (arguments == null || arguments.isEmpty() || arguments.size() < 3) {
```
jdneo: `arguments.isEmpty()` & `arguments.size() < 3`. Duplicated check?

testforstephen: nice catch.

PR #299, opened 2019-09-19
Add minor installation note in README.md about mvnw.cmd/mvnw

p10q:
Just an idea. Probably there's a better way to do this. But for those getting started with Java again (and not as familiar with maven wrapper script), it did take a minute to figure out what to do to get it installed (i.e., that `mvn clean install` doesn't work).


README.md
```diff
@@ -4,6 +4,17 @@
...
+
+### Windows:
+```
+./mvnw.cmd clean install
+```
+### Linux and macOS:
+```
+./mvnw clean install
+```
+
```
testforstephen: How about moving it behind the `Repository Structure` paragraph?

p10q: Cool yeah! Sounds good.

PR #302, opened 2019-10-12
Check whether java file is on classpath

testforstephen:


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaDebugDelegateCommandHandler.java
```diff
@@ -97,4 +102,19 @@ public Object executeCommand(String commandId, List<Object> arguments, IProgress
 
         return result;
     }
+
+    private boolean isOnClasspath(List<Object> arguments) {
+        if (arguments.size() < 1) {
+            return true;
```
Eskibear: false?

testforstephen: i prefer to throw an exception.

Eskibear: agree, and it would be easier to find out any future abuse of the command (if there is)

PR #305, opened 2019-10-22
Support searching main classes from the workspace invisible project

testforstephen:


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainClassHandler.java
```diff
@@ -141,6 +144,15 @@ public void acceptSearchMatch(SearchMatch match) {
         return resolutions;
     }
 
+    private boolean isContainedInInvisibleProject(IProject project, Collection<IPath> rootPaths) {
+        if (project == null) {
+            return false;
+        }
+
+        IFolder workspaceLinkFolder = project.getFolder("_");
```
jdneo: Use `org.eclipse.jdt.ls.core.internal.ProjectUtils.WORKSPACE_LINK` to replace `"_"`

testforstephen: resolved

PR #325, opened 2020-03-27
Fix race competition when the breakpoint needs be evaluated in multi threads

testforstephen:

Fix microsoft/vscode-java-debug#782

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Watchpoint.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Watchpoint.java
@@ -14,6 +14,7 @@ package com.microsoft.java.debug.core;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 
@@ -21,9 +22,11 @@ import org.apache.commons.lang3.StringUtils;
 
 import com.sun.jdi.Field;
 import com.sun.jdi.ReferenceType;
+import com.sun.jdi.ThreadReference;
 import com.sun.jdi.VMDisconnectedException;
 import com.sun.jdi.VirtualMachine;
 import com.sun.jdi.event.ClassPrepareEvent;
+import com.sun.jdi.event.ThreadDeathEvent;
 import com.sun.jdi.request.ClassPrepareRequest;
 import com.sun.jdi.request.EventRequest;
 import com.sun.jdi.request.WatchpointRequest;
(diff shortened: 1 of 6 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Watchpoint.java
```diff
@@ -155,7 +168,7 @@ public Object getProperty(Object key) {
         requests.add(classPrepareRequest);
 
         CompletableFuture<IWatchpoint> future = new CompletableFuture<>();
-        Disposable subscription = eventHub.events()
+        subscription = eventHub.events()
```
andxu, line 171: add subscriptions.add(subscription)?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Watchpoint.java
```diff
@@ -249,4 +262,14 @@ public void setCompiledLogpointExpression(Object compiledExpression) {
...
     }
+
+    @Override
+    public Object getCompiledExpression(long threadId) {
+        return compiledExpressions.get(threadId);
+    }
+
+    @Override
+    public void setCompiledExpression(long threadId, Object compiledExpression) {
+        compiledExpressions.put(threadId, compiledExpression);
```
andxu, line 273: put/get in multiple thread is error prone.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
```diff
@@ -148,20 +148,12 @@ public void initialize(IDebugAdapterContext context, Map<String, Object> props)
             ASTEvaluationEngine engine = new ASTEvaluationEngine(project, debugTarget);
             boolean newExpression = false;
             if (breakpoint != null) {
-                if (StringUtils.isNotBlank(breakpoint.getLogMessage())) {
```
andxu, line 151: why this condition is deleted: StringUtils.isNotBlank(breakpoint.getLogMessage())?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Watchpoint.java
testforstephen: the following L182 already did that.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/eval/JdtEvaluationProvider.java
testforstephen: this is not needed any more. Previously the compiled expression for the logpoint and conditional breakpoint are saved separately, so when evaluating a breakpoint, need check whether it's a logpoint or conditional breakpoint. In the new implementation, they are saved in a map, and fetched from the map directly.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Watchpoint.java
testforstephen: nice catch!

PR #333, opened 2020-05-11
Fix gradle test scope

testforstephen:

Fix microsoft/vscode-java-debug#715


testforstephen:
For gradle project, the upstream buildship would check the main type to determine whether to include the test scope.

https://github.com/eclipse/buildship/blob/d1353a31ca863be2f38cabaf3a9c8cec34598c2e/org.eclipse.buildship.core/src/main/java/org/eclipse/buildship/core/internal/launch/LaunchConfigurationScope.java#L57

PR #335, opened 2020-06-22
Improve the HCR workflow

testforstephen:


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JavaHotCodeReplaceProvider.java
```diff
@@ -300,14 +303,21 @@ public void onClassRedefined(Consumer<List<String>> consumer) {
...
+            boolean success;
             synchronized (this) {
                 classNames.addAll(deltaClassNames);
-                doHotCodeReplace(deltaResources, deltaClassNames);
+                success = doHotCodeReplace(deltaResources, deltaClassNames);
                 deltaResources.clear();
                 deltaClassNames.clear();
             }
+
+            if (!classNames.isEmpty() && !success) {
```
Eskibear, line 317: What if `classNames` is empty? (I imagine it's the case of clicking HCR without any code changes, repeatedly). Does it return successfully without a warning?

testforstephen: The client will show a message saying no changed classes.

PR #336, opened 2020-06-23
Display the method return value when stepping out

testforstephen:

Close microsoft/vscode-java-debug#660


testforstephen:
image

jdneo:
What if the method has parameters? And what will it looks like when it's a invocation chain (like stream invocation)

Eskibear:
With this icon '⎯►', then why you still need to append "returned", for better clarification?

testforstephen:
Addressed, below is the new look.
image

PR #344, opened 2020-07-30
Migrate to Java 11

testforstephen:


testforstephen:
Due to the issue https://github.com/redhat-developer/vscode-java/issues/1551, cannot build java-debug with the vscode-java extension.

PR #353, opened 2020-09-23
Use the correct runtime to validate the JVM versions between the debugger and debuggee

testforstephen:
When attaching to an existing debuggee program, its runtime version may be different with the one that the debugger used to build your project. Although the debugger still works when the JDK mismatches. But stepping into the JDK system library, the native library source line may be not consistent with the remote one and cause the line mismatch during stepping. For such case, the debugger will print a warning message when runtime is not matched in the DEBUG CONSOLE view.

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
@@ -20,29 +20,12 @@ import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
-import org.apache.commons.lang3.StringUtils;
-import org.eclipse.core.resources.IResource;
-import org.eclipse.core.runtime.CoreException;
-import org.eclipse.debug.core.sourcelookup.ISourceContainer;
-import org.eclipse.jdt.core.IBuffer;
-import org.eclipse.jdt.core.IClassFile;
-import org.eclipse.jdt.core.IJavaElement;
-import org.eclipse.jdt.core.IJavaProject;
-import org.eclipse.jdt.core.ITypeRoot;
-import org.eclipse.jdt.core.JavaCore;
-import org.eclipse.jdt.core.JavaModelException;
-import org.eclipse.jdt.core.dom.AST;
-import org.eclipse.jdt.core.dom.ASTParser;
-import org.eclipse.jdt.core.dom.CompilationUnit;
-import org.eclipse.jdt.internal.debug.core.breakpoints.ValidBreakpointLocationLocator;
 
 import com.microsoft.java.debug.core.Configuration;
 import com.microsoft.java.debug.core.DebugException;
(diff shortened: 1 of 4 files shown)


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -181,6 +187,37 @@ public String getSourceFileURI(String fullyQualifiedName, String sourcePath) {
         return null;
     }
 
+    private static final Pattern JAVA_VERSION_PATTERN = Pattern.compile("version \"(.*)\"");
```
Eskibear, line 190: Just FYI that the pattern **could** not cover some corner case (though we haven't figure out why), see https://github.com/AdoptOpenJDK/openjdk-build/issues/1610

Eskibear (dismissed):

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
testforstephen: Thanks for the info. I plan to reuse the JDT logic to get the Java version, where it uses `release` file under the Java home to get the version. I assume it's more stable regarding the jdt user base.

```java
	public synchronized String readReleaseVersion(File javaHome) {

		String version = ""; //$NON-NLS-1$

		if (Files.notExists(Paths.get(javaHome.getAbsolutePath(), RELEASE_FILE))) {
			return version;
		}
		try (Stream<String> lines = Files.lines(Paths.get(javaHome.getAbsolutePath(), RELEASE_FILE), Charset.defaultCharset()).filter(s -> s.contains(JAVA_VERSION))) {
			Optional<String> hasVersion = lines.findFirst();
			if (hasVersion.isPresent()) {
				String line = hasVersion.get();
				version = line.substring(14, line.length() - 1); // length of JAVA_VERSION + 2 in JAVA_VERSION="9"
			}
		}
		catch (UncheckedIOException | IOException e) {
			LaunchingPlugin.log(e);
		}

		return version;

	}
```

jdneo: @CsCherrYY FYI

Eskibear: @testforstephen `release` file looks great! I tried several, **one distribution of AdoptOpenJDK uses major version** while others use full version. But it's the same value with `java -version` output.

Zulu:
```
IMPLEMENTOR="Azul Systems, Inc."
IMPLEMENTOR_VERSION="Zulu14.28+21-CA"
JAVA_VERSION="14.0.1"
```

Adopt:
```
IMPLEMENTOR="AdoptOpenJDK"
IMPLEMENTOR_VERSION="AdoptOpenJDK"
JAVA_VERSION="14"    // <-------------HERE
```
```
IMPLEMENTOR="AdoptOpenJDK"
IMPLEMENTOR_VERSION="AdoptOpenJDK"
JAVA_VERSION="13.0.2"
```

Oracle:
```
BUILD_TYPE="commercial"
IMPLEMENTOR="Oracle Corporation"
JAVA_VERSION="12.0.2"
```

jdneo (approved):

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
testforstephen: YES. In real world, maybe 99% cases will contain `release` file. You could find the Java version from the release file first. If it doesn't exist, then fall back to `java -version` command.

PR #354, opened 2020-10-10
Expose a command to resolve the source mapping for the specified stacktrace

testforstephen:


jdneo (approved):

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveSourceMappingHandler.java
```diff
@@ -0,0 +1,53 @@
...
+        if (matcher.find()) {
+            String methodField = matcher.group(2);
+            String locationField = matcher.group(matcher.groupCount());
+            String fullyQualifiedName = methodField.substring(0, methodField.lastIndexOf("."));
+            String packageName = fullyQualifiedName.lastIndexOf(".") > -1 ? fullyQualifiedName.substring(0, fullyQualifiedName.lastIndexOf(".")) : "";
+            String[] locations = locationField.split(":");
+            String sourceName = locations[0];
+            String sourcePath = StringUtils.isBlank(packageName) ? sourceName
+                    : packageName.replace('.', File.separatorChar) + File.separatorChar + sourceName;
+            return sourceProvider.getSourceFileURI(fullyQualifiedName, sourcePath);
```
puremourning, line 48: I think this needs to apply user source mappings like the code in StackTraceRequestHandler does.

That way, the additional source paths can be used to find the actual source locations rather than jdt:// stuff.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -178,14 +178,7 @@ public String getSourceFileURI(String fullyQualifiedName, String sourcePath) {
...
         } else if (sourceElement instanceof IClassFile) {
-            try {
-                IClassFile file = (IClassFile) sourceElement;
-                if (file.getBuffer() != null) {
-                    return getFileURI(file);
-                }
-            } catch (JavaModelException e) {
-                // do nothing.
-            }
+            return getFileURI((IClassFile) sourceElement);
```
puremourning, line 181: This change breaks most debug adapter clients finding the source.

The effect of this change is that getSourceFileURI always returns an (arguably invalid) jdt:// path now (as this method always returns a value), and so the code in StackTraceRequestHandler no longer actually checks user-specified source locations.

It's always better to use user-specified source locations (that contain the _actual source_) than returning a jdt uri which requires another server round-trip and somehow the server to (hopefully) find the source associated with the .class file.

puremourning (commented):
I have just bisected a break in my DAP client that traced back to this.

Previously stack traces were returning paths based on the `sourcePaths` entries in the launch configuration, but after an upgrade to java-debug, this stopped working and all paths were returned (when remote debugging) as `jdt://` paths. These are not handled by my client, so all remote debugging broke.

I tracked it down to this change which always returns the "jdt://" path now, even if the file is not resolved. I reverted the changes to JdtSourceLookUpProvider.java and it all works again.

I admit I'm not very familiar with these APIs, but it seems clear to me that the check on buffer() != null was making these source directories work.

The code in stack trace handler basically uses the user's source paths if the lookup provider returns null.

PR #362, opened 2021-02-23
Ensure CompletionsResponse contains targets

mfussenegger:
If a client sent a completion request without a frameId then the server
returned a completion response without targets.

According to the specification the response of a completions request
must always contain a `targets` property. It is not optional.

See https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Completions

---

I couldn't find any existing test cases for completions, am I missing something?


mfussenegger:
Any inputs on this?

testforstephen (approved):

testforstephen:
thank you for the contribution.

PR #367, opened 2021-04-10
modify project order in SourceContainers for increase project priority

mozhuanzuojing:
- Because some project dependencies may have to have some jars of source code, and in general, we give priority to debugging the project source code


testforstephen:
Could you please open an issue first and share a sample project to reproduce this issue?

mozhuanzuojing:
After two years of use, no significant difference was found. So I will close this branch.

PR #368, opened 2021-04-12
Leverage AST and Java Runtime to resolve inline values

testforstephen:
The steps:

 1. Visit AST tree to find the variable tokens in each lines.
 
    ASTVisitor can identify if a node is a variable declaration and a simple name. And then based on resolveBinding, it's easy to tell if the name is a variable/field binding. 

 2. Resolve the values of the variable tokens.
 
    Once the variable names in each lines are found, next step is to resolve values for these variables. Usually the variables come from four places, we can use different methods to get the values.
    - Local variables declared in current method body
    - Formal parameters declared in method declaration.
    - Field variables of the enclosing instance.
    - Captured variables from outer scope. For example, local type is accessing variables of enclosing method, and lambda expression body is accessing to variables of enclosing method.

    For the first two kinds, they are visible local variables of the current stack frame, and by default their values will be expanded by Variables View. So we're going to return them with VariableLookup kind to tell client to look them up in Variables View directly.

    For the last two kinds, their values are hidden as properties of 'this' object and require additional evaluation to get their values. So we're going to return them with Evaluation kind, and then calling the debuggee runtime to evaluate their values.


testforstephen:
The variables from outer scope are captured in current stack frame.

image

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/InlineValuesRequestHanlder.java
```diff
@@ -0,0 +1,250 @@
...
+import com.sun.jdi.IntegerValue;
+import com.sun.jdi.Method;
+import com.sun.jdi.ObjectReference;
+import com.sun.jdi.ReferenceType;
+import com.sun.jdi.StackFrame;
+import com.sun.jdi.Value;
+
+import org.apache.commons.lang3.math.NumberUtils;
+
+public class InlineValuesRequestHanlder implements IDebugRequestHandler {
```
jdneo, line 55: typo: Hanlder

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/InlineValuesRequestHanlder.java
```diff
@@ -0,0 +1,250 @@
...
+                    InlineVariable inlineVariable = unresolvedVariables[i];
+                    boolean isInlineFieldVariable = (inlineVariable.declaringClass != null);
+                    boolean isMatch = false;
+                    if (isSyntheticLambdaFrame) {
+                        isMatch = !isInlineFieldVariable && Objects.equals(fieldName, inlineVariable.expression);
+                    } else {
+                        boolean isMatchedField = isInlineFieldVariable
+                            && Objects.equals(fieldName, inlineVariable.expression)
+                            && Objects.equals(typeName, inlineVariable.declaringClass);
+                        boolean isMathcedCapturedVariable = !isInlineFieldVariable
```
jdneo, line 217: typo: Mathc

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/InlineValueHandler.java
```diff
@@ -0,0 +1,654 @@
...
+            VariableVisitor visitor = new VariableVisitor(astRoot, stoppedRange, params.viewPort, Flags.isStatic(enclosingMethod.getFlags()));
+            astRoot.accept(visitor);
+            InlineVariable[] result = visitor.getInlineVariables();
+            return result;
+        } catch (JavaModelException e) {
+            return new InlineVariable[0];
+        }
+    }
+
+    private static IMethod findEnclosingMethod(ITypeRoot root, int stoppedOffset) throws JavaModelException {
```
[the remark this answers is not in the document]
testforstephen: No, it doesn't apply to this task. The purpose of that helper is to find the definition element of the selection. For example, the offset is at the `String▮`, it returns the String.class element. But what we want is the enclosing method `method`.

```java
void method() {
    String▮ a = "test";
}

```

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/InlineValuesRequestHanlder.java
testforstephen: nice catch!

jdneo (approved):

PR #369, opened 2021-04-21
Refresh variables with new variable formatters

testforstephen:


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Requests.java
```diff
@@ -372,6 +380,7 @@ public static RunInTerminalRequestArguments createExternalTerminal(String[] cmds
         CONTINUEOTHERS("continueOthers", ThreadOperationArguments.class),
         PAUSEALL("pauseAll", ThreadOperationArguments.class),
         PAUSEOTHERS("pauseOthers", ThreadOperationArguments.class),
+        REFRESHVARIABLES("refreshVariables", RefreshVariablesArguments.class),
```
Eskibear, line 383: `REFRESH_VARIABLES` would be better human-readable. For consistency just leave it as is. But consider to refactor all the  enums one day?

Eskibear (approved):

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Requests.java
testforstephen: yes, i will use a separate PR to refactor these enums.

Eskibear (approved):

PR #383, opened 2021-09-13
Allow mainClass to be configured with unicode characters

testforstephen:
'SourceVersion.isName' is a utility method from JDK class, it can be used to validate if a class name is a valid qualified name.

Fixes microsoft/vscode-java-debug#780

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainClassHandler.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainClassHandler.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017 Microsoft Corporation and others.
+ * Copyright (c) 2017-2021 Microsoft Corporation and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/ResolveMainClassHandler.java
```diff
@@ -185,23 +189,41 @@ private ValidationResponse validateLaunchConfigCore(List<Object> arguments) thro
...
+                    CONFIGERROR_MAIN_CLASS_NOT_UNIQUE);
             }
         }
 
         return new ValidationResult(true);
     }
 
+    // Java command line supports two kinds of main class format: <mainclass> and <module>[/<mainclass>]
+    private boolean isValidMainClassName(String mainClass) {
+        if (StringUtils.isEmpty(mainClass)) {
```
jdneo, line 214: Curious what will happen if main class name is empty?

testforstephen: when main class is not specified (e.g. empty), the validator will pass. because the debugger will auto resolve them from search engine once it's not configured by user.

jdneo: I see..

jdneo (approved):

PR #385, opened 2021-09-22
if listing the variable value of a large object fails, just display its basic info

testforstephen:
Fixes microsoft/vscode-java-debug#1044

Since JDI does not provide an efficient way to query whether a variable is a large object or not. One way to mitigate this is to intercept OOM or timeout exceptions that occur when fetching variable information and display a placeholder on these failed variable segments. This way, the user still has the opportunity to view the basic variable information.

**Before** - Variables view just displays an Exception message.
image

**After** - Variables view still lists all variables, but displays a placeholder for the failed variable property.
image


jdneo (commented):
Overall looks good to me. Just curious what's the behavior of Eclipse/IntelliJ for such case?

testforstephen:
> Overall looks good to me. Just curious what's the behavior of Eclipse/IntelliJ for such case?

Nice concern.

In eclipse, it shows a dialog about OOM and meanwhile the variable view shows an empty row for the exceptional variable.

image

image

In IJ, it uses async to display each variable and there is a never-ending loading message "Collecting data...".
image

jdneo (approved):

PR #391, opened 2021-11-18
Make DAP implementation to be encoding-neutral

testforstephen:
This requires https://github.com/microsoft/vscode-java-debug/pull/1077

This project is an implementation of DAP, it's better to keep encoding neutral. It's the client's responsibility to determine what encoding to use to launch user's application.

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2017-2020 Microsoft Corporation and others.
+ * Copyright (c) 2017-2021 Microsoft Corporation and others.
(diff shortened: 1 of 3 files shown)


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -120,11 +119,7 @@ public boolean supportsRealtimeBreakpointVerification() {
         String filePath = AdapterUtils.toPath(uri);
         // For file uri, read the file contents directly and pass them to the ast parser.
         if (filePath != null && Files.isRegularFile(Paths.get(filePath))) {
-            Charset cs = (Charset) this.options.get(Constants.DEBUGGEE_ENCODING);
-            if (cs == null) {
-                cs = Charset.defaultCharset();
-            }
-            String source = readFile(filePath, cs);
+            String source = readFile(filePath);
```
testforstephen, line 122: When parsing Java source files, let's keep using the same encoding as JDT language server.

jdneo (approved):

Eskibear (approved):

PR #401, opened 2022-03-14
Support lazy loading object values from toString()

CsCherrYY:
requires: https://github.com/microsoft/vscode-java-debug/pull/1134

specification: https://microsoft.github.io/debug-adapter-protocol/specification#Types_VariablePresentationHint

demo:

lazy

more details can be found in https://github.com/microsoft/vscode/issues/135147, and there is still a UX discussion.

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/VariablesRequestHandler.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/VariablesRequestHandler.java
@@ -96,7 +96,7 @@ public class VariablesRequestHandler implements IDebugRequestHandler {
 
         VariableProxy containerNode = (VariableProxy) container;
 
-        if (containerNode.isLazyResolved() && DebugSettings.getCurrent().showToString) {
+        if (containerNode.isLazyVariable() && DebugSettings.getCurrent().showToString) {
             Types.Variable typedVariable = this.resolveLazyVariable(context, containerNode, variableFormatter, options, evaluationEngine);
             if (typedVariable != null) {
                 list.add(typedVariable);
@@ -276,11 +276,9 @@ public class VariablesRequestHandler implements IDebugRequestHandler {
                 }
             }
 
-            int referenceId = 0;
             VariableProxy varProxy = null;
             if (indexedVariables > 0 || (indexedVariables < 0 && value instanceof ObjectReference)) {
                 varProxy = new VariableProxy(containerNode.getThread(), containerNode.getScope(), value, containerNode, evaluateName);
-                referenceId = context.getRecyclableIdPool().addObject(containerNode.getThreadId(), varProxy);
                 varProxy.setIndexedVariable(indexedVariables >= 0);
                 varProxy.setUnboundedType(javaVariable.isUnboundedType());
             }
@@ -307,9 +305,6 @@ public class VariablesRequestHandler implements IDebugRequestHandler {
                 typeString = "";
             }
 
-            Types.Variable typedVariables = new Types.Variable(name, valueString, typeString, referenceId, evaluateName);
-            typedVariables.indexedVariables = Math.max(indexedVariables, 0);
-
             String detailsValue = null;
             if (hasErrors) {
                 // If failed to resolve the variable value, skip the details info as well.
@@ -317,8 +312,7 @@ public class VariablesRequestHandler implements IDebugRequestHandler {
                 detailsValue = "size=" + variableFormatter.valueToString(sizeValue, options);
             } else if (DebugSettings.getCurrent().showToString) {
                 if (VariableDetailUtils.isLazyLoadingSupported(value) && varProxy != null) {
-                    typedVariables.presentationHint = new VariablePresentationHint(true);
-                    varProxy.setLazyResolved(true);
+                    varProxy.setLazyVariable(true);
                 } else {
                     try {
                         detailsValue = VariableDetailUtils.formatDetailsValue(value, containerNode.getThread(), variableFormatter, options, evaluationEngine);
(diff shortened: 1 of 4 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/VariablesRequestHandler.java
```diff
@@ -94,6 +95,24 @@
         }
 
         VariableProxy containerNode = (VariableProxy) container;
+
+        if (containerNode.getReferencedVariableId() != null && DebugSettings.getCurrent().showToString) {
```
testforstephen, line 99: I'd suggest to extract the following block to a private method like `resolveLazyValue(variable)` for more readable.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/variables/VariableProxy.java
```diff
@@ -109,4 +110,12 @@ public boolean isUnboundedType() {
     public void setUnboundedType(boolean isUnboundedType) {
         this.isUnboundedType = isUnboundedType;
     }
+
+    public Integer getReferencedVariableId() {
```
testforstephen, line 114: I suggest to add a property `isLazyResolved()` instead of `getReferencedVariableId`. Its default value is `false`. And set it to `true` when you return that variable as lazy presentation.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/VariablesRequestHandler.java
```diff
@@ -305,14 +323,21 @@
...
-                    logger.log(Level.SEVERE, "Failed to compute the toString() value of a large object", e);
-                    detailsValue = "<Unable to display the details of a large object>";
-                } catch (Exception e) {
-                    logger.log(Level.SEVERE, "Failed to compute the toString() value", e);
-                    detailsValue = "<Failed to resolve the variable details due to \"" + e.getMessage() + "\">";
+                if (VariableDetailUtils.isLazyLoadingSupported(value) && varProxy != null) {
+                    typedVariables.presentationHint = new VariablePresentationHint(true);
+                    VariableProxy valueReferenceProxy = this.getVariableProxy(containerNode, value, "", indexedVariables, javaVariable);
+                    Integer referencedVariableId = context.getRecyclableIdPool().addObject(containerNode.getThreadId(), valueReferenceProxy);
+                    varProxy.setReferencedVariableId(referencedVariableId);
```
testforstephen, line 330: No need to regenerate a VariableProxy, just use a new property `isLazyResolved` to mark this variable to be lazy resolved.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/VariablesRequestHandler.java
CsCherrYY: fixed in [c8366d5](https://github.com/microsoft/java-debug/pull/401/commits/c8366d51f901ca3fae535aadda0f4c75b2354bb7)

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/VariablesRequestHandler.java
```diff
@@ -331,6 +347,24 @@
...
+        valueReferenceProxy.setIndexedVariable(containerNode.isIndexedVariable());
+        valueReferenceProxy.setUnboundedType(containerNode.isUnboundedType());
+        int referenceId = context.getRecyclableIdPool().addObject(containerNode.getThreadId(), valueReferenceProxy);
+        Object proxiedVariable = containerNode.getProxiedVariable();
+        if (proxiedVariable instanceof ObjectReference) {
+            ObjectReference variable = (ObjectReference) proxiedVariable;
+            String valueString = variableFormatter.valueToString(variable, options);
+            String detailString = VariableDetailUtils.formatDetailsValue(variable, containerNode.getThread(), variableFormatter, options,
+                    evaluationEngine);
+            return new Types.Variable("", valueString + " " + detailString, "", referenceId, "");
```
testforstephen, line 363: When running it on an old vscode, you will find variable name info is missing.

image

testforstephen (approved):
the latest commit works.

beckermarc:
Is it somehow possible to disable this? I now have to do a lot of clicks to view the desired information during debugging...

CsCherrYY:
@beckermarc you can set configuration `java.debug.settings.showToString` to false to disable this.

beckermarc:
Thanks. However that seems to disable the whole feature. I'd prefer to only disable the lazy loading. Is that also possible?

CsCherrYY:
> Thanks. However that seems to disable the whole feature. I'd prefer to only disable the lazy loading. Is that also possible?

@testforstephen any comments?

testforstephen:
Make sense to have a setting to turn it off.

beckermarc:
I think my main concern with this change is that it also seems to hide more information than just the toString() value:

Here is one example before the version that introduced this change:
Before

Here is the same variable with the new version. Note that I also don't see the class anymore and are no longer made aware of the fact that I can expand this variable further:
After (not loaded)

After clicking the ellipsis I get the same information as before:
After (loaded)

Especially when trying to get an overview over all variables in the current scope this requires me to do a lot more clicks. I never ran into issues with long running toString() methods in my projects so far, so it would be great to be able to disable this from my perspective.

testforstephen:
yes, it would be better to display the class name during lazy loading. This is an improvement we also reported back to VS Code core team. https://github.com/microsoft/vscode/issues/135147#issuecomment-1062758849

beckermarc:
cool, thanks for your work!

veastark:
Please include an option to disable the lazy loading and return to the old method of displaying variable values. With this addition debugging has become very painful.

testforstephen:
VS Code has added a setting `debug.autoExpandLazyVariables` to control whether to auto expand lazy variable. That will be released in next VS Code version. Currently you can try it with VS Code Insider first.

PR #411, opened 2022-04-20
fix evaluation variable

mozhuanzuojing:
Background: The company project has been reporting errors in the evaluation of variables in the parent class. 
After this modification test, there is no problem, and the evaluation variables are successful. Although I don't particularly understand why


mozhuanzuojing:
Since debugging the plugin at the company, the problem doesn't appear anymore, it's amazing. Temporarily close this commit.

PR #412, opened 2022-04-29
Add documentation for the supported launch/attach properties

mfussenegger:
Useful for users of clients other than vscode


mfussenegger:
Just realised that the options are already documented in https://github.com/microsoft/vscode-java-debug#options

I guess this can be closed,  unless you find it worth including in this repo as well

testforstephen:
I'm closing this PR. We put all docs at vscode-java-debug repo because it's the repo that the end user will use.

PR #413, opened 2022-05-22
Capture Java process pid and display debug toolbar when running Java in terminal

testforstephen:
It does two things in this PR:
- Capture the Java process pid when running Java in a terminal shell.
   This is a follow-up of https://github.com/microsoft/java-debug/pull/399. Both the SpringBoot Dashboard and Spring Tools extension have the requirement to get the running Java pid from Java debugger, so it makes sense to support it in Java debugger directly for reuse.

   Since the DAP just returns the shell process pid when running Java in a terminal shell, the debugger have to find the running Java process based on the shell pid. In the standard shell environment such as PowerShell, cmd, Linux/macOS shells, the shell process is parent of Java process, it's easy to find Java subprocess from process tree. But this approach fails when using Git Bash as the default terminal, because the launched Java process is no longer a descendant process of the shell process.

  Git bash is based on MSYS2 platform, which is a fork of Cygwin technology. When running Cygwin exe in Windows, it runs at an emulator layer. So its process tree is not same as Windows process tree. Each Cygwin process has a mapped Windows pid. Running ps command in bash can list the cygwin's pid, cygwin's ppid and its winpid. Based on ps command, it's possible to find the running Java process and its parent Git bash process.

- Display the debug toolbar when using "Run Java" action to run Java in terminal without debugging.
   Keep the debug session alive until the Java process exits. This fixes microsoft/vscode-java-debug#1166
   image

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/DisconnectRequestWithoutDebuggingHandler.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/DisconnectRequestWithoutDebuggingHandler.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2018 Microsoft Corporation and others.
+* Copyright (c) 2018-2022 Microsoft Corporation and others.
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * which accompanies this distribution, and is available at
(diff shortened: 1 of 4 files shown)


Eskibear (approved):
LGTM

testforstephen:
With this PR, there are two ways to get the running Java process pid.
- Custom DAP request `processId`.
- `processid` notification.

And the response/event body will contain
```
{
 shellProcessId: number  /* The process ID of the terminal shell if the process is running in a terminal shell.  */
 processId: number /* The process ID. */
}
```

PR #414, opened 2022-05-24
Naming Java terminal with mainClass name

testforstephen:
Fixes microsoft/vscode-java-debug#1164

Naming the terminal for "Run Java" action as "Run \<MainClassName\>". For "Debug Java" action, naming it as "Debug \<MainClassName\>". The same debug session will reuse the same terminal. And it can also differentiate the Run and Debug behavior.

image


jdneo:
How about adding a `:` between the 'mode' and the 'main class name'? Like `Run: Hello`, `Debug: Hello`?

testforstephen:
> How about adding a `:` between the 'mode' and the 'main class name'? Like `Run: Hello`, `Debug: Hello`?

It looks good with a separator `:`.

image

PR #424, opened 2022-06-27
Support method breakpoints

testforstephen:
Add support for function breakpoint for java. The current version support adding a function breakpoint with the following syntax.

`FullyQualifiedClassName#methodName`

like

`java.util.ArrayList#add`

The current version doesn't consider parameter signatures.

 This PR is based on PR https://github.com/microsoft/java-debug/pull/418, and addresses the additional issues as follows:
- Handle the duplicated MethodEntryEvent better.
- Update the reason of BreakpontEvent to be 'changed'.
- Fix checkstyle issues.

Thanks to the initial contribution from @gayanper.


jdneo:
Overall, the new functionality works well. One concern is about the inner class case.

Since we require binary name here, so user has to use `$` as the separator. Is there any way to give users some hints when `.` is misused in that case?

testforstephen:
> Overall, the new functionality works well. One concern is about the inner class case.
> 
> Since we require binary name here, so user has to use `$` as the separator. Is there any way to give users some hints when `.` is misused in that case?

Agreed, currently adding a function breakpoint from the BREAKPOINTS panel is very limited and we need a more user friendly way to guide the user in enanling function breakpoints.

As a first release, I will keep the current implementation. I opened an issue in microsoft/vscode-java-debug#1183 and hope to improve function breakpoints in the next release.

jdneo (approved):

PR #427, opened 2022-07-20
Add support for lambda breakpoints

gayanper:
The support for method header breakpoints was extended to support
lambda breakpoints using the vscode inline breakpoint feature.

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
@@ -38,10 +38,21 @@ public class LambdaExpressionLocator extends ASTVisitor {
             int startPosition = node.getStartPosition();
 
             int startColumn = this.compilationUnit.getColumnNumber(startPosition);
-            int endColumn = this.compilationUnit.getColumnNumber(startPosition + node.getLength());
-            int lineNumber = this.compilationUnit.getLineNumber(startPosition);
+            int endPosition = startPosition + node.getLength();
+            int endColumn = this.compilationUnit.getColumnNumber(endPosition);
+            int startLine = this.compilationUnit.getLineNumber(startPosition);
+            int endLine = this.compilationUnit.getLineNumber(endPosition);
 
-            if (column >= startColumn && column <= endColumn && lineNumber == line) {
+            // lambda on same line:
+            // list.stream().map(i -> i + 1);
+            //
+            // lambda on multiple lines:
+            // list.stream().map(user
+            // -> user.isSystem() ? new SystemUser(user) : new EndUser(user));
+
+            if ((startLine == endLine && column >= startColumn && column <= endColumn && line == startLine)
+                    || (startLine != endLine && line >= startLine && line <= endLine
+                            && (column >= startColumn || column <= endColumn))) {
                 this.lambdaMethodBinding = node.resolveMethodBinding();
                 this.found = true;
                 this.lambdaExpression = node;
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
@@ -34,15 +34,13 @@ public class LambdaExpressionLocator extends ASTVisitor {
 
     @Override
     public boolean visit(LambdaExpression node) {
-        if (column > -1) {
+        // we only support inline breakpoints which are added before the expression part of the
+        // lambda. And we don't support lambda blocks since they can be debugged using line
+        // breakpoints.
+        if (column > -1 && node.getNodeType() != ASTNode.BLOCK) {
             int startPosition = node.getStartPosition();
-
-            int startColumn = this.compilationUnit.getColumnNumber(startPosition);
-            int endPosition = startPosition + node.getLength();
-            int endColumn = this.compilationUnit.getColumnNumber(endPosition);
-            int startLine = this.compilationUnit.getLineNumber(startPosition);
-            int endLine = this.compilationUnit.getLineNumber(endPosition);
-
+            int endPosition = node.getBody().getStartPosition();
+            int offset = this.compilationUnit.getPosition(line, column);
             // lambda on same line:
             // list.stream().map(i -> i + 1);
             //
@@ -50,9 +48,7 @@ public class LambdaExpressionLocator extends ASTVisitor {
             // list.stream().map(user
             // -> user.isSystem() ? new SystemUser(user) : new EndUser(user));
 
-            if ((startLine == endLine && column >= startColumn && column <= endColumn && line == startLine)
-                    || (startLine != endLine && line >= startLine && line <= endLine
-                            && (column >= startColumn || column <= endColumn))) {
+            if (offset >= startPosition && offset <= endPosition) {
                 this.lambdaMethodBinding = node.resolveMethodBinding();
                 this.found = true;
                 this.lambdaExpression = node;
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
@@ -37,10 +37,11 @@ public class LambdaExpressionLocator extends ASTVisitor {
         if (column > -1) {
             int startPosition = node.getStartPosition();
 
-            int columnNumber = this.compilationUnit.getColumnNumber(startPosition);
+            int startColumn = this.compilationUnit.getColumnNumber(startPosition);
+            int endColumn = this.compilationUnit.getColumnNumber(startPosition + node.getLength());
             int lineNumber = this.compilationUnit.getLineNumber(startPosition);
 
-            if (column == columnNumber && lineNumber == line) {
+            if (column >= startColumn && column <= endColumn && lineNumber == line) {
                 this.lambdaMethodBinding = node.resolveMethodBinding();
                 this.found = true;
                 this.lambdaExpression = node;
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
@@ -1,0 +1,97 @@
+/*******************************************************************************
+* Copyright (c) 2022 Microsoft Corporation and others.
+* All rights reserved. This program and the accompanying materials
+* are made available under the terms of the Eclipse Public License v1.0
+* which accompanies this distribution, and is available at
+* http://www.eclipse.org/legal/epl-v10.html
+*
+* Contributors:
+*     Gayan Perera (gayanper@gmail.com) - initial API and implementation
+*******************************************************************************/
+package com.microsoft.java.debug;
+
+import org.eclipse.jdt.core.dom.ASTNode;
+import org.eclipse.jdt.core.dom.ASTVisitor;
+import org.eclipse.jdt.core.dom.AbstractTypeDeclaration;
+import org.eclipse.jdt.core.dom.CompilationUnit;
+import org.eclipse.jdt.core.dom.IMethodBinding;
+import org.eclipse.jdt.core.dom.LambdaExpression;
+
+public class LambdaExpressionLocator extends ASTVisitor {
+    private CompilationUnit compilationUnit;
+    private int line;
+    private int column;
+    private boolean found;
+
+    private IMethodBinding lambdaMethodBinding;
+    private LambdaExpression lambdaExpression;
+
+    public LambdaExpressionLocator(CompilationUnit compilationUnit, int line, int column) {
+        this.compilationUnit = compilationUnit;
+        this.line = line;
+        this.column = column;
+    }
+
+    @Override
+    public boolean visit(LambdaExpression node) {
+        if (column > -1) {
+            int startPosition = node.getStartPosition();
+
+            int columnNumber = this.compilationUnit.getColumnNumber(startPosition);
+            int lineNumber = this.compilationUnit.getLineNumber(startPosition);
+
+            if (column == columnNumber && lineNumber == line) {
+                this.lambdaMethodBinding = node.resolveMethodBinding();
+                this.found = true;
+                this.lambdaExpression = node;
+                return false;
+            }
+        }
+        return super.visit(node);
+    }
+
+    /**
+     * Returns <code>true</code> if a lambda is found at given location.
+     */
+    public boolean isFound() {
+        return found;
+    }
+
+    /**
+     * Returns the signature of lambda method otherwise return null.
+     */
+    public String getMethodSignature() {
+        if (!this.found) {
+            return null;
+        }
+        return BreakpointLocationLocator.toSignature(this.lambdaMethodBinding, getMethodName());
+    }
+
+    /**
+     * Returns the name of lambda method otherwise return null.
+     */
+    public String getMethodName() {
+        if (!this.found) {
+            return null;
+        }
+        String key = this.lambdaMethodBinding.getKey();
+        return key.substring(key.indexOf('.') + 1, key.indexOf('('));
+    }
+
+    /**
+     * Returns the name of the type which the lambda method is found.
+     */
+    public String getFullyQualifiedTypeName() {
+        if (this.found) {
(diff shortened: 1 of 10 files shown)


gayanper:
@testforstephen let me know what you think about this change. May be the name of the breakpoint displayed in UI might need a change. I just named it as lambda breakpoint.

gayanper:
By the way I had to add a new locator implementation to find LambdaExpressions since the JDT.Core doesn't work out of the box since it works with offsets

testforstephen:
@gayanper Is there any code snippet to demo how this works?

gayanper:
> @gayanper Is there any code snippet to demo how this works?

https://user-images.githubusercontent.com/7471994/180408840-55184075-e601-4a84-99ea-17535f87740c.mp4

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -155,23 +156,42 @@ public String[] getFullyQualifiedName(String uri, int[] lines, int[] columns) th
...
-                // When the final valid line location is same as the original line, that represents it's a valid breakpoint.
-                // Add location type check to avoid breakpoint on method/field which will never be hit in current implementation.
-                if (lines[i] == locator.getLineLocation()
-                        && locator.getLocationType() == BreakpointLocationLocator.LOCATION_LINE) {
-                    fqns[i] = locator.getFullyQualifiedTypeName();
-                } else if (locator.getLocationType() == BreakpointLocationLocator.LOCATION_METHOD) {
-                    fqns[i] = locator.getFullyQualifiedTypeName().concat("#").concat(locator.getMethodName())
-                            .concat("#").concat(locator.getMethodSignature());
+                // if we have a column, try to find the lambda expression at that column
+                LambdaExpressionLocator lambdaExpressionLocator = new LambdaExpressionLocator(astUnit, lines[i],
```
testforstephen, line 160: If it's not an inline breakpoint (e.g. column number is -1), we can just skip the LambdaVisitor here.

testforstephen (commented):
I tried adding an inline breakpoint on a lambda, it works well with current implementation. This is a good start for inline breakpoint support. But I'm not sure whether current approach can be extended to other inline breakpoint case. I'm thinking whether there is a systematic way to handle inline breakpoint.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
gayanper: True my bad, I was planning to skip the unwanted visiting. I will fix it.

gayanper:
> I tried adding an inline breakpoint on a lambda, it works well with current implementation. This is a good start for inline breakpoint support. But I'm not sure whether current approach can be extended to other inline breakpoint case. I'm thinking whether there is a systematic way to handle inline breakpoint.

I think we can add support for that as well. The same class we use to search for Lambda expression can be evolve into finding MethodInvocations at AST level at the given column and create the breakpoint for that particular method like we use for method header breakpoints and lambda. I think i can work on it if you think that will add value. I have not come across java IDEs provide such a feature. So may be it will be really good one to have.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public boolean visit(LambdaExpression node) {
+        if (column > -1) {
+            int startPosition = node.getStartPosition();
+
+            int columnNumber = this.compilationUnit.getColumnNumber(startPosition);
+            int lineNumber = this.compilationUnit.getLineNumber(startPosition);
+
+            if (column == columnNumber && lineNumber == line) {
```
testforstephen, line 43: Currently the lambda inline breakpoint must be set at the beginning of the lambda expression, which is not friendly for lambda that uses parentheses to wrap args.

`list.stream().forEach(▮(name) -> System.out.println(name))` works, but 
`list.stream().forEach((▮name) -> System.out.println(name))` does not.

I'm wondering whether we should relax the lambda breakpoint condition, e.g.  
the column breakpoint is valid as long as it stops within the lambda parameter list.

@gayanper WDYT?

testforstephen:
> I think we can add support for that as well. The same class we use to search for Lambda expression can be evolve into finding MethodInvocations at AST level at the given column and create the breakpoint for that particular method like we use for method header breakpoints and lambda. I think i can work on it if you think that will add value. I have not come across java IDEs provide such a feature. So may be it will be really good one to have.

We can release the lambda inline breakpoint first. Currently, inline breakpoints are not so obvious to new users. I believe one day we need to support BreakpointLocationsRequest https://github.com/microsoft/vscode-java-debug/issues/1193, which will hint user the possible inline breakpoints in a line.

gayanper:
> > I think we can add support for that as well. The same class we use to search for Lambda expression can be evolve into finding MethodInvocations at AST level at the given column and create the breakpoint for that particular method like we use for method header breakpoints and lambda. I think i can work on it if you think that will add value. I have not come across java IDEs provide such a feature. So may be it will be really good one to have.
> 
> We can release the lambda inline breakpoint first. Currently, inline breakpoints are not so obvious to new users. I believe one day we need to support BreakpointLocationsRequest [microsoft/vscode-java-debug#1193](https://github.com/microsoft/vscode-java-debug/issues/1193), which will hint user the possible inline breakpoints in a line.

Thats on my radar as well ;)

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
gayanper: Yes i think that make more sense, I can update the criteria to support it.

gayanper: Try now @testforstephen

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public boolean visit(LambdaExpression node) {
+        if (column > -1) {
+            int startPosition = node.getStartPosition();
+
+            int startColumn = this.compilationUnit.getColumnNumber(startPosition);
+            int endColumn = this.compilationUnit.getColumnNumber(startPosition + node.getLength());
+            int lineNumber = this.compilationUnit.getLineNumber(startPosition);
+
+            if (column >= startColumn && startColumn <= endColumn && lineNumber == line) {
```
testforstephen, line 44: The algorithm is not accurate.

For the snippet `list.stream().map(name -> name).forEach((▮item) -> System.out.println(item))`, it always matches the first lambda `name -> name`.

testforstephen (commented):
Found two bugs with the snippet `list.stream().map(▮name -> name).forEach((▮item) -> System.out.println(item))`:
- If adding 2+ inline breakpoints in the same line, only one is added successfully.
- The second inline breakpoint is always recognized as `lambda$1`, but it's supposed to be `lambda$2`.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
gayanper: I think the issue is with line number.  I will fix it today evening and test the scenarios.

testforstephen:
> Found two bugs with the snippet `list.stream().map(▮name -> name).forEach((▮item) -> System.out.println(item))`:
> 
> * If adding 2+ inline breakpoints in the same line, only one is added successfully.
> * The second inline breakpoint is always recognized as `lambda$1`, but it's supposed to be `lambda$2`.

The first issue still exists. The reason is current BreakpointManager uses LineNumber to check if breakpoints are the same. For the inline breakpoints on the same line, the BreakpointManager will add them as the same inline breakpoint.
https://github.com/microsoft/java-debug/blob/8c0e87defedc5bc2a17a9e2980824f8573c08386/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/BreakpointManager.java#L80-L84

And the optimized lambda search works for single line lambdas, but does not work for multi-line lambdas. For mult-line lambda, one idea is to limit to the available range to the lambda start position to the body start position.
```
list.stream().map(🔴name -> name).forEach((🔴item) -> {
    System.out.println(item);
});
```

gayanper:
@testforstephen i will look at this in the evening. Did a quick correction in the morning when i saw that i was using wrong column variable. But i think for multi line spanning we need another fix

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
gayanper: I was able to fix the multi line issue and I think i can commit the both fixes including multiple lambda on same line by end of tomorrow

gayanper:
@testforstephen i was wondering if we need to send back the column to client as well ?

testforstephen:
> @testforstephen i was wondering if we need to send back the column to client as well ?

Are there any clients that can consume the returned column info?  If so, it would be good to return the actual column covered by the breakpoint.

gayanper:
The protocol has the support and I thought to have the column in the internal breakpoin to make it unique and use line:column as the map key. Now since we have the column we can send it back. 

Not sure if vscode will use it for anything or not.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
gayanper: i have now fixed and tested for both multiple lambda's on same line and lambdas spanning multiple lines.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
```diff
@@ -0,0 +1,109 @@
...
+            int endColumn = this.compilationUnit.getColumnNumber(endPosition);
+            int startLine = this.compilationUnit.getLineNumber(startPosition);
+            int endLine = this.compilationUnit.getLineNumber(endPosition);
+
+            // lambda on same line:
+            // list.stream().map(i -> i + 1);
+            //
+            // lambda on multiple lines:
+            // list.stream().map(user
+            // -> user.isSystem() ? new SystemUser(user) : new EndUser(user));
```
testforstephen, line 51: Do we need to consider the block-style lambda?

```java
list.stream().map(🔴user -> {
    System.out.println(user);
    return user.isSystem() ? new SystemUser(user) : new EndUser(user))
});
```

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
```diff
@@ -0,0 +1,109 @@
...
+
+            if ((startLine == endLine && column >= startColumn && column <= endColumn && line == startLine)
+                    || (startLine != endLine && line >= startLine && line <= endLine
+                            && (column >= startColumn || column <= endColumn))) {
```
testforstephen, line 55: this condition only works for expression-style lambda with 2 lines.

gayanper: I tested for

```
        Arrays.asList(1, 2, 3).stream().map(name -> name).forEach(🔴(item) -> {
            System.out.println(item + 1 + 2 + 3 + 4 + 5);
            System.out.println(item + 1 + 2 + 3 + 4 + 5);
            System.out.println(item + 1 + 2 + 3 + 4 + 5);
            System.out.println(item + 1 + 2 + 3 + 4 + 5);
        });
```

and

```
        Arrays.asList(1, 2, 3).stream().map(name -> name).forEach(
            (item) -> 🔴{
            System.out.println(item + 1 + 2 + 3 + 4 + 5);
            System.out.println(item + 1 + 2 + 3 + 4 + 5);
            System.out.println(item + 1 + 2 + 3 + 4 + 5);
            System.out.println(item + 1 + 2 + 3 + 4 + 5);
        });
```

Its working, can you give a example snippet ?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
gayanper: Yes i tested it as i have mentioned below

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
testforstephen: Case 1:
image
This will hit lambda entry on line 21.

Case 2:
image
This will hit lambda entry on line 22.

> (startLine != endLine && line >= startLine && line <= endLine
                            && (column >= startColumn || column <= endColumn))

It is not accurate to compare the column number for multi-line lambda. It might be better to compare the offset.

gayanper: True. But I couldn’t find a method to calculate offset by providing the line and column. But will check more to see if i have missed any such method. JDT should have such methods, not sure if they are part of the API.

testforstephen: You can use `this.compilationUnit#getPosition(line, column)` to get the offset.

gayanper: Thanks. I think i missed it. The problem with current condition is I didn’t intend to support blocks. But still if we only do boundry check and ignore middle lines it will work when we have multi lines. But position is much simpler. I will switch to that.

gayanper: For the scenario 1 as user what you would expect ?

testforstephen: > For the scenario 1 as user what you would expect ?

I expect adding a breakpoint on (Line 22, Column 24), and if Column 24 breakpoint is not supported, then fallback to a line breakpoint on Line 22. 

If it's inside a block body, I'll treat it as a normal function body and not fall back to the Lambda entry breakpoint.

testforstephen: LambdaExpression.getBody() will return either a Block or an Expression, which we can use to determine which lambda it is.

If it's an expression-style lambda, it's ok to enable Lambda breakpoint if the offset is within the whole LambdaExpression range.
If it's a block-style lambda, I would enable Lambda breakpoint only if the offset is in [LambdaStart, LambdaBodyStart).

Just my two cents.

gayanper: Now keeping mind the future expansion of inline breakpoints for method invocations, I decided to restrict the lambda breakpoints by
- only support for expressions
- the inline breakpoint should be before the expression

This way we will not break any functionality nor complicate breakpoint logic in future when we introduce method invocation inline breakpoints. And for lambda blocks we can always use line breakpoints which is much clear and straight forward.

WDYT ?

testforstephen (approved):
I agree with simplifying the condition check of lambda breakpoint. Checking for the offset within [lambdaStart, bodyStart] works well for me.

One minor comment is to remove the unncessary nodeType check. Since LambdaExpression.getNodeType() returns `ASTNode.LAMBDA_EXPRESSION`, the condition `node.getNodeType() != ASTNode.BLOCK` is always true and i just remove it from the commit.

Others look good to me.

testforstephen:
Had a try on the typescript inline breakpoint, it would show an extra cursor on the target column when hitting on inline breakpoint. It's a better experience we can leverage.

image

Damian558:
> The support for method header breakpoints was extended to support
> lambda breakpoints using the vscode inline breakpoint feature.

Damian_siner@hotmail.com

PR #428, opened 2022-07-20
Show target VM exceptions as result in evaluate requests

mfussenegger:
This changes the result of an expression like `Long.parseLong("foo")`
sent to the evaluate handler from:

    org.eclipse.debug.core.DebugException: com.sun.jdi.InvocationException: Exception occurred in target VM occurred invoking method.

to the actual exception:

    NumberFormatException@76 "java.lang.NumberFormatException: For input string: "foo""
      backtrace: Object[6]@82
      cause: NumberFormatException@76
      depth: 46
      detailMessage: "For input string: "foo""
      stackTrace: StackTraceElement[0]@84
      suppressedExceptions: Collections$EmptyList@85 size=0

I think this makes the console in vscode and other DAP clients more useful as you get a more meaningful result for expressions that fail


testforstephen (approved):
@mfussenegger It's a nice improvement. Thanks for contribution.

PR #429, opened 2022-07-24
Support Show as Binary

mylog00:
Hi there!
This is PR for allow represent variables in binary forma. 
I think it may be helpful for vscode-java-debug issue [#1078](https://github.com/microsoft/vscode-java-debug/issues/1078)
I've added new protocol parameter 'formatType' should contains variable format such as [BIN, OCT, HEX, DEC].
I've remain "showHex" for backward compatibility and can be delete later when PR to vscode-java-debug will be merged.
If you are not interested in this changes fill free to close this PR :)

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Requests.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Requests.java
@@ -24,7 +24,9 @@ import com.microsoft.java.debug.core.protocol.Types.DataBreakpoint;
 public class Requests {
 
     public static class ValueFormat {
+        @Deprecated
         public boolean hex;
+        public String type;
     }
 
     public static class Arguments {
@@ -300,7 +302,9 @@ public class Requests {
     public static class RefreshVariablesArguments extends Arguments {
         public boolean showStaticVariables = false;
         public boolean showQualifiedNames = false;
+        @Deprecated
         public boolean showHex = false;
+        public String formatType = null;
         public boolean showLogicalStructure = true;
         public boolean showToString = true;
     }
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,6 @@ hs_err_pid*
 target/
 .settings
 **/.checkstyle
-target/
 bin/
 **/lib/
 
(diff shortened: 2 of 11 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Requests.java
```diff
@@ -24,7 +24,9 @@
 public class Requests {
 
     public static class ValueFormat {
+        @Deprecated
         public boolean hex;
```
mfussenegger, line 28: `hex` is part of the specification: https://microsoft.github.io/debug-adapter-protocol/specification#Types_ValueFormat, what's the rationale for deprecating this?

mylog00: Because of new functionality we can format values to different formats now, beside hex. And I added new property `ValueFormat#type` instead of `ValueFormat#hex` to support that possibility. I assumed we can delete `ValueFormat#hex` in future with no harm because of no need and marked `hex` as deprecated. Just tell me if I did something wrong and I'll fix it. :)

testforstephen: @mylog00 The data types under the folder `com/microsoft/java/debug/core/protocol` are generated from the [DAP](https://microsoft.github.io/debug-adapter-protocol/specification) spec, we can't just change it on debugger side.

It's OK to extend the debug settings to support more format types, but remember to support the old valueFormat.hex as well.

testforstephen: When contributing to the DAP, you need to open an issue to discuss the proposal first. Only if the proposal is accepted, then open a PR.

A quick search and found a duplicated issue for more formatting options. https://github.com/microsoft/debug-adapter-protocol/issues/197

mylog00: Got it. Thanks)

mylog00: Hi! If there is no other issue could you approve this PR?)

PR #433, opened 2022-08-04
Remove inline lambda breakpoints for blocks

gayanper:
Remove the possibility to add inline lambda breakpoints for lambda blocks
since it cause confusion with paused line and breakpoint line.


gayanper:
@testforstephen update the condition to remove inline breakpoint possibility from blocks.

testforstephen:
Currently the available offset is limited to the range of [lambdaStart, bodyStart], this applies to all kinds of Lambda Expression. I think it's not necessary to exclude the block-style lambdas.

gayanper:
The reason is as you can see in the following picture, the line doesn't sync with breakpoint location
image

testforstephen:
> @testforstephen The reason is as you can see in the following picture, the line doesn't sync with breakpoint location image

This is also true for function breakpoints. Add a breakpoint on the method header , but the actual stop line is the first valid line of the body. So, I think it's ok.

PR #445, opened 2022-09-19
Visualize the inline breakpoint locations

testforstephen:
What this PR does:
- Support BreakpointLocations DAP Request, this can visualize the locations where inline breakpoints can be added. Closes microsoft/vscode-java-debug#1193.
image

- Show an extra column cursor when an inline breakpoint is hit. Closes microsoft/vscode-java-debug#1202
image

- Line breakpoint will only break on the first location on that line. If the line contains other lambdas, you can add an inline breakpoint to break on other locations.

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
@@ -37,13 +37,11 @@ import io.reactivex.disposables.Disposable;
 public class Breakpoint implements IBreakpoint {
     private VirtualMachine vm = null;
     private IEventHub eventHub = null;
-    private String className = null;
-    private int lineNumber = 0;
+    private JavaBreakpointLocation sourceLocation = null;
     private int hitCount = 0;
     private String condition = null;
     private String logMessage = null;
     private HashMap<Object, Object> propertyMap = new HashMap<>();
-    private String methodSignature = null;
 
     private boolean async = false;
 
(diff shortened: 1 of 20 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
```diff
@@ -149,6 +173,7 @@ public void setHitCount(int hitCount) {
             .filter(request -> request instanceof BreakpointRequest)
             .subscribe(request -> {
                 request.addCountFilter(hitCount);
+                request.disable();
```
jdneo, line 176: why we need to call `disable` first here?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -181,24 +229,56 @@ public String[] getFullyQualifiedName(String uri, int[] lines, int[] columns) th
...
+                    sourceLocation.setMethodSignature(locator.getMethodSignature());
                 }
             }
         }
-        return fqns;
+
+        return sourceLocations;
+    }
+
+    private BreakpointLocation[] getInlineBreakpointLocations(final CompilationUnit astUnit, int sourceLine) {
```
jdneo, line 260: Besides the lambda expression. Do we support multiple statements in one line? for example: 

```java
int a = 0; int b = 1; int c = a+b;
```

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
testforstephen: image

Calling request.enable() will do nothing if the request has been enabled before. We need to invalidate the old request, and re-enable it to send the new JDWP request to debuggee.

https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/96cda748fde8d874fff6e3fad1f3f775377eac93/org.eclipse.jdt.debug/jdi/org/eclipse/jdi/internal/request/EventRequestImpl.java#L245-L248

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
testforstephen: We only support adding an inline breakpoint for lambda only. This is also the behavior of IJ.

jdneo (approved):

PR #459, opened 2022-11-17
Only report build failure when the build errors are on the running project's classpath

testforstephen:
- For a build-tool managed project, will check if the running project and its references projects contain build errors. If so, report build failure. Otherwise, return succeed.
- For unmanaged folder, only check if the running main class contains build errors. If so, report build failure. Otherwise, return succeed.

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/Compile.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/Compile.java
@@ -1,0 +1,145 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Microsoft Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


jdneo:
I'm thinking that how we can re-use the code in https://github.com/eclipse/eclipse.jdt.ls/blob/d4101ae14289ff7db98763cf731a0c74138589b8/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/BuildWorkspaceHandler.java#L100

Now we build the entire workspace but only check the markers for the trigger project. That means we may waste some time for building unrelated projects.

testforstephen:
> I'm thinking that how we can re-use the code in https://github.com/eclipse/eclipse.jdt.ls/blob/d4101ae14289ff7db98763cf731a0c74138589b8/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/handlers/BuildWorkspaceHandler.java#L100
> 
> Now we build the entire workspace but only check the markers for the trigger project. That means we may waste some time for building unrelated projects.

I will hold on current behavior first before the performance becomes a concern.

jdneo (approved):

Issue #477, opened 2023-04-04
NPE when adding lambda breakpoints in reactive projects


gayanper:
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.core.dom.IMethodBinding.getKey()" because "binding" is null
	at com.microsoft.java.debug.BindingUtils.getMethodName(BindingUtils.java:36)
	at com.microsoft.java.debug.LambdaExpressionLocator.getMethodName(LambdaExpressionLocator.java:84)
	at com.microsoft.java.debug.plugin.internal.JdtSourceLookUpProvider.getBreakpointLocations(JdtSourceLookUpProvider.java:176)
	at com.microsoft.java.debug.core.adapter.handler.SetBreakpointsRequestHandler.convertClientBreakpointsToDebugger(SetBreakpointsRequestHandler.java:312)
	at com.microsoft.java.debug.core.adapter.handler.SetBreakpointsRequestHandler.handle(SetBreakpointsRequestHandler.java:108)
	at com.microsoft.java.debug.core.adapter.DebugAdapter.lambda$dispatchRequest$0(DebugAdapter.java:93)
	at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1187)
	at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2309)
	at com.microsoft.java.debug.core.adapter.DebugAdapter.dispatchRequest(DebugAdapter.java:92)
	at com.microsoft.java.debug.core.adapter.ProtocolServer.dispatchRequest(ProtocolServer.java:118)
	at com.microsoft.java.debug.core.protocol.AbstractProtocolServer.lambda$new$0(AbstractProtocolServer.java:81)
	at io.reactivex.internal.observers.LambdaObserver.onNext(LambdaObserver.java:63)
	at io.reactivex.internal.operators.observable.ObservableObserveOn$ObserveOnObserver.drainNormal(ObservableObserveOn.java:201)
	at io.reactivex.internal.operators.observable.ObservableObserveOn$ObserveOnObserver.run(ObservableObserveOn.java:255)
	at io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:66)
	at io.reactivex.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:57)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[...]

PR #478, opened 2023-04-04
Improve lambda method discovery Fixes: #477

gayanper:
The issue seems to be when resolving lambda method for reactive operators, the binding resolver is not the default binding resolvers which actually resolves the lambda method. 

It seems. the AST construction effects which binding resolver is used at runtime in the node. Therefore switched to the using cache AST when available which seems to resolved the issue. Will do some more testing.

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/BindingUtils.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/BindingUtils.java
@@ -35,7 +35,15 @@ public final class BindingUtils {
      */
     public static String getMethodName(IMethodBinding binding, boolean fromKey) {
         if (fromKey) {
-            return LambdaLocationLocatorHelper.toMethodName(binding);
+            String key = binding.getKey();
+            int dotAt = key.indexOf('.');
+            int end = key.indexOf('<', dotAt);
+            if (end == -1) {
+                end = key.indexOf('(');
+            } else {
+                end = Math.min(end, key.indexOf('('));
+            }
+            return key.substring(dotAt + 1, end);
         } else {
             return binding.getName();
         }
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/BindingUtils.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/BindingUtils.java
@@ -12,10 +12,12 @@
 package com.microsoft.java.debug;
 
 import org.eclipse.jdt.core.dom.IMethodBinding;
+import org.eclipse.jdt.internal.debug.core.breakpoints.LambdaLocationLocatorHelper;
 
 /**
  * Utility methods around working with JDT Bindings.
  */
+@SuppressWarnings("restriction")
 public final class BindingUtils {
     private BindingUtils() {
 
@@ -33,46 +35,21 @@ public final class BindingUtils {
      */
     public static String getMethodName(IMethodBinding binding, boolean fromKey) {
         if (fromKey) {
-            String key = binding.getKey();
-            int dotAt = key.indexOf('.');
-            int end = key.indexOf('<', dotAt);
-            if (end == -1) {
-                end = key.indexOf('(');
-            } else {
-                end = Math.min(end, key.indexOf('('));
-            }
-            return key.substring(dotAt + 1, end);
+            return LambdaLocationLocatorHelper.toMethodName(binding);
         } else {
             return binding.getName();
         }
     }
 
     /**
-     * Returns the method signature of the method represented by the binding. Since
-     * this implementation use the {@link IMethodBinding#getKey()} to extract the
-     * signature from, the method name must be passed in.
+     * Returns the method signature of the method represented by the binding
+     * including the synthetic outer locals.
      *
      * @param binding the binding which the signature must be resolved for.
-     * @param name    the name of the method.
-     * @return the signature or null if the signature could not be resolved from the
-     *         key.
+     * @return the signature or null if the signature could not be resolved.
      */
-    public static String toSignature(IMethodBinding binding, String name) {
-        // use key for now until JDT core provides a public API for this.
-        // "Ljava/util/Arrays;.asList<T:Ljava/lang/Object;>([TT;)Ljava/util/List<TT;>;"
-        // "([Ljava/lang/String;)V|Ljava/lang/InterruptedException;"
-        String signatureString = binding.getKey();
-        if (signatureString != null) {
-            name = "." + name;
-            int index = signatureString.indexOf(name);
-            if (index > -1) {
-                int exceptionIndex = signatureString.indexOf("|", signatureString.lastIndexOf(")"));
-                if (exceptionIndex > -1) {
-                    return signatureString.substring(index + name.length(), exceptionIndex);
-                }
-                return signatureString.substring(index + name.length());
-            }
-        }
-        return null;
+    public static String toSignature(IMethodBinding binding) {
+        return LambdaLocationLocatorHelper.toMethodSignature(binding);
     }
+
 }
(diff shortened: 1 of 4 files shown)


testforstephen:
https://github.com/microsoft/java-debug/blob/31cc03389a7890aa8ff5be6c897ca2575b910329/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java#L266-L280

When parsing the AST for the lambda, we just passed a char array without the context Java model or project, that's why the binding information is not resolved correctly.

Your workaround seems OK. Another way to fix this is to find the Java project that the current file belongs to and set the parser to use that project for binding resolver.

gayanper:
@testforstephen WDYT about the latest change ?

gayanper:
@testforstephen anything pending for approval and merge ?

testforstephen:
this PR is on the list for May sprint.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/BindingUtils.java
```diff
@@ -33,46 +35,21 @@ private BindingUtils() {
...
-            String key = binding.getKey();
-            int dotAt = key.indexOf('.');
-            int end = key.indexOf('<', dotAt);
-            if (end == -1) {
-                end = key.indexOf('(');
-            } else {
-                end = Math.min(end, key.indexOf('('));
-            }
-            return key.substring(dotAt + 1, end);
+            return LambdaLocationLocatorHelper.toMethodName(binding);
```
testforstephen, line 38: Is there any case that the helper covers better?

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/BindingUtils.java
```diff
@@ -33,46 +35,21 @@ private BindingUtils() {
...
-                int exceptionIndex = signatureString.indexOf("|", signatureString.lastIndexOf(")"));
-                if (exceptionIndex > -1) {
-                    return signatureString.substring(index + name.length(), exceptionIndex);
-                }
-                return signatureString.substring(index + name.length());
-            }
-        }
-        return null;
+    public static String toSignature(IMethodBinding binding) {
+        return LambdaLocationLocatorHelper.toMethodSignature(binding);
```
testforstephen, line 52: Same for this. Is there any case covered better by the helper?

testforstephen (commented):
> https://github.com/microsoft/java-debug/blob/31cc03389a7890aa8ff5be6c897ca2575b910329/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java#L266-L280
> 
> When parsing the AST for the lambda, we just passed a char array without the context Java model or project, that's why the binding information is not resolved correctly.

A simpler fix is to use the parent Java project as the environment. This works for the case in issue #477.
```java
            IFile resource = (IFile) JDTUtils.findResource(JDTUtils.toURI(uri), ResourcesPlugin.getWorkspace().getRoot()::findFilesForLocationURI);
            if (resource != null && JdtUtils.isJavaProject(resource.getProject())) {
                parser.setProject(JavaCore.create(resource.getProject()));
            } else {
                parser.setEnvironment(new String[0], new String[0], null, true);
            }
```

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/BindingUtils.java
gayanper: Yes it handles the method resolutions better without depending on the binding key if I remember correctly. 

I know its a internal class, but i guess rather duplicating the same logic we could look at making it public if we want to avoid internal api usage in future. 

WDYT ?

gayanper:
> > https://github.com/microsoft/java-debug/blob/31cc03389a7890aa8ff5be6c897ca2575b910329/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java#L266-L280
> > 
> > When parsing the AST for the lambda, we just passed a char array without the context Java model or project, that's why the binding information is not resolved correctly.
> 
> A simpler fix is to use the parent Java project as the environment. This works for the case in issue #477.
> 
> ```java
>             IFile resource = (IFile) JDTUtils.findResource(JDTUtils.toURI(uri), ResourcesPlugin.getWorkspace().getRoot()::findFilesForLocationURI);
>             if (resource != null && JdtUtils.isJavaProject(resource.getProject())) {
>                 parser.setProject(JavaCore.create(resource.getProject()));
>             } else {
>                 parser.setEnvironment(new String[0], new String[0], null, true);
>             }
> ```

This is a good suggestion, I change the code now.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -276,8 +278,14 @@ private CompilationUnit asCompilationUnit(String uri) {
...
              */
-            parser.setEnvironment(new String[0], new String[0], null, true);
-            parser.setUnitName(Paths.get(filePath).getFileName().toString());
+            IFile resource = (IFile) JDTUtils.findResource(JDTUtils.toURI(uri),
+                    ResourcesPlugin.getWorkspace().getRoot()::findFilesForLocationURI);
+            if (resource != null && JdtUtils.isJavaProject(resource.getProject())) {
+                parser.setProject(JavaCore.create(resource.getProject()));
+            } else {
+                parser.setEnvironment(new String[0], new String[0], null, true);
+                parser.setUnitName(Paths.get(filePath).getFileName().toString());
```
testforstephen, line 287: `parser.setUnitName(Paths.get(filePath).getFileName().toString());` is required for the case of Java project configured as well.

testforstephen: There is NPE without the unit name:
```
!MESSAGE [error response][setBreakpoints]: Cannot invoke "org.eclipse.jdt.core.dom.ITypeBinding.getBinaryName()" because the return value of "org.eclipse.jdt.core.dom.AbstractTypeDeclaration.resolveBinding()" is null
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.core.dom.ITypeBinding.getBinaryName()" because the return value of "org.eclipse.jdt.core.dom.AbstractTypeDeclaration.resolveBinding()" is null
	at com.microsoft.java.debug.LambdaExpressionLocator.getFullyQualifiedTypeName(LambdaExpressionLocator.java:96)
	at com.microsoft.java.debug.plugin.internal.JdtSourceLookUpProvider.getBreakpointLocations(JdtSourceLookUpProvider.java:177)
	at com.microsoft.java.debug.core.adapter.handler.SetBreakpointsRequestHandler.convertClientBreakpointsToDebugger(SetBreakpointsRequestHandler.java:312)
	at com.microsoft.java.debug.core.adapter.handler.SetBreakpointsRequestHandler.handle(SetBreakpointsRequestHandler.java:108)
```

gayanper: Fixed

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/BindingUtils.java
testforstephen: Actually the implementation here is a little different with the helper method.

`Git blame` shows that it's this PR https://github.com/microsoft/java-debug/commit/69230b339d68dfa2b3a1c251c29f3bd577fd0cf8#diff-91a75ad1996749ac4383bad13f73aaab20a962d82966486c06a09bb6dd86d5ff that changed the implementation a little bit. Do you remember the history why you changed it last time?

gayanper: Its for handling keys which contains generics. basically strip the generics parts from the method name.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/BindingUtils.java
gayanper: The JDT one handles synthetic outer variables as well properly.

PR #491, opened 2023-05-25
Improve support for multiline lambda expressions. Fix #489

### LINKED ISSUE #489: Google formatter cause inline lambda issue
Take the following example

```
  @org.junit.jupiter.api.Test
  void testMain1() {
    System.out.println();
    List<String> list =
        List.of("A", "1").stream().filter(
          a -> "1".equals(a)
        ).collect(Collectors.toList());

    assertEquals(1, list.size());
  }
```

I cannot add an inline breakpoint to `a -> "1".equals(a)`, because it is considered as a line breakpoint. But at the same time line breakpoint doesn't check if this line is under a lambda and make it a lambda breakpoint. If I use a lambda block things works or format it like

```
  @org.junit.jupiter.api.Test
  void testMain1() {
    System.out.println();
    List<String> list =
        List.of("A", "1").stream().filter(a
          -> "1".equals(a)
        ).collect(Collectors.toList());

    assertEquals(1, list.size());
  }
```

gayanper:
The patch tries to see if the line breakpoint falls on a lambda expression which is wrapped into a new line by the formatter.
Therefore when trying to see if the current line falls inside a lambda we use a range match instead of a exact match of the start position. For this purpose a new constructor parameter is added into Lambda Locator class to switch between this mode.

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/LambdaExpressionLocator.java
@@ -49,7 +49,17 @@ public class LambdaExpressionLocator extends ASTVisitor {
             // Since the debugger supports BreakpointLocations Request to hint user
             // about possible inline breakpoint locations, we will only support
             // inline breakpoints added where a lambda expression begins.
-            if (breakOffset == startPosition || (!exactMatch && breakOffset > startPosition && breakOffset < (startPosition + node.getLength()))) {
+
+            // the second expression tries to find lambda expression which are formatted as
+            // list.stream().map(
+            // user -> user.isSystem() ? new SystemUser(user) : new EndUser(user));
+            // since lambda is on a different line it will not match the first expression.
+
+            if (breakOffset == startPosition
+                    || (!exactMatch
+                            && this.compilationUnit.getLineNumber(breakOffset) == this.compilationUnit
+                                    .getLineNumber(startPosition)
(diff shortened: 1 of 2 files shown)


gayanper:
I didn't check the hint, i think i might have to use the same logic for hints as well. But adding from context menu seems to not straight forward isn't it only for this particular case, even though its a workaround.

testforstephen:
Can you explain how does this fix improve the user experience?

gayanper:
> Can you explain how does this fix improve the user experience?

For example as a user who doesn't know how java DAP works, i will be tempted to add a breakpoint at the line where the lambda expression is and expect to pause. Just for this case using context menu might not be really intuitive for a user isn't it ?

I also found the right click operation still doesn't work if there are nested lambdas ? Also my current solution doesn't support it either as of now. 

But i can look i to improving this.

gayanper:
@testforstephen can you check the latest changes ? It should be working as expected now.

testforstephen:
```java
  @org.junit.jupiter.api.Test
  public void testMain1() {
    System.out.println();
    List<String> list =
        List.of("A", "1").stream().filter(
          a -> "1".equals(a)).collect(Collectors.toList()); // add breakpoint at this line

    assertEquals(1, list.size());
  }
```

The case above still doesn't work.

gayanper:
@testforstephen i think in this scenario we have a valid line breakpoint which matches, therefore we don't search for lambda locations. 

I suspect with the way we have implemented line breakpoints which we only focus on the line, not other locations present like lambdas, we have to handle this as you mentioned using right click menu. 

WDYT?

PR #495, opened 2023-06-26
Debug support on the decompiled source code

testforstephen:
this depends on redhat.java@1.20.2023062004 and later to make it work.

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
@@ -36,6 +36,7 @@ import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.URIUtil;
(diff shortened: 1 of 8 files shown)


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -533,4 +571,58 @@ private boolean isSameURI(String uri1, String uri2) {
...
+            }
+
+            if (classFile.getSourceRange() == null) {
+                ContentProviderManager contentProvider = JavaLanguageServerPlugin.getContentProviderManager();
+                try {
+                    DecompilerResult result = contentProvider.getSourceResult(classFile, new NullProgressMonitor());
+                    if (result != null) {
+                        return result.getOriginalLineMappings();
+                    }
+                } catch (NoSuchMethodError e) {
```
jdneo, line 594: Will the `NoClassDefFoundError` throw here? I remember `DecompilerResult` is newly added as well.

testforstephen: > DecompilerResult result = contentProvider.getSourceResult(classFile, new NullProgressMonitor());

`getSourceResult` is also a new method, and will be thrown first.

jdneo: ok

jdneo (approved):

PR #499, opened 2023-07-01
Fix generic return type lambda breakpoint issue. Fix #1359

gayanper:
The fix tries to compare the runtime lambda method signature with none generic version of the method signature found from AST traversal as a additional comparison to what is there.

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
@@ -343,14 +343,14 @@ public class Breakpoint implements IBreakpoint {
         for (int i = 0; i < chars.length; i++) {
             char c = chars[i];
             if (c == '<') {
-                depth ++;
+                depth++;
                 append = (depth == 0);
             }
             if (append) {
                 builder.append(c);
             }
             if (c == '>') {
-                depth --;
+                depth--;
                 append = (depth == 0);
             }
         }
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
@@ -335,24 +335,23 @@ public class Breakpoint implements IBreakpoint {
         return location;
     }
 
-    private String toNoneGeneric(String genericSig) {
+    static String toNoneGeneric(String genericSig) {
         StringBuilder builder = new StringBuilder();
-        boolean skip = false;
+        boolean append = true;
+        int depth = 0;
         char[] chars = genericSig.toCharArray();
         for (int i = 0; i < chars.length; i++) {
             char c = chars[i];
             if (c == '<') {
-                skip = true;
+                depth ++;
+                append = (depth == 0);
             }
-            if (!skip) {
+            if (append) {
                 builder.append(c);
             }
-            if (skip && c == '>') {
-                if (i + 2 < chars.length) {
-                    skip = chars[i + 2] == '>';
-                } else {
-                    skip = false;
-                }
+            if (c == '>') {
+                depth --;
+                append = (depth == 0);
             }
         }
         return builder.toString();
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
@@ -338,7 +338,9 @@ public class Breakpoint implements IBreakpoint {
     private String toNoneGeneric(String genericSig) {
         StringBuilder builder = new StringBuilder();
         boolean skip = false;
-        for (char c : genericSig.toCharArray()) {
+        char[] chars = genericSig.toCharArray();
+        for (int i = 0; i < chars.length; i++) {
+            char c = chars[i];
             if (c == '<') {
                 skip = true;
             }
@@ -346,7 +348,11 @@ public class Breakpoint implements IBreakpoint {
                 builder.append(c);
             }
             if (skip && c == '>') {
-                skip = false;
+                if (i + 2 < chars.length) {
+                    skip = chars[i + 2] == '>';
+                } else {
+                    skip = false;
+                }
             }
         }
         return builder.toString();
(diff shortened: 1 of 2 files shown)


gayanper:
@testforstephen Let me know what you think about this change. This is vita when debugging reactive projects, specially when setting lambda breakpoints at Mono.flatMap operations.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
```diff
@@ -325,7 +325,8 @@ private Location findMethodLocaiton(ReferenceType refType, String methodName, St
         for (Method method : methods) {
             if (!method.isAbstract() && !method.isNative()
                     && methodName.equals(method.name())
-                    && (methodSiguature.equals(method.genericSignature()) || methodSiguature.equals(method.signature()))) {
+                    && (methodSiguature.equals(method.genericSignature()) || methodSiguature.equals(method.signature())
+                            || toNoneGeneric(methodSiguature).equals(method.signature()))) {
```
testforstephen, line 329: `toNoneGeneric()` function is duplicated with `org.eclipse.jdt.core.Signature.getTypeErasure()`, we can reuse the one from the JDT core directly.

testforstephen: This is debug core code, no dependency on JDT. Anyway, current implementation of `toNoneGeneric` has bugs, we can copy the implementation of `org.eclipse.jdt.core.Signature.getTypeErasure()`.

Here is a sample.
```java
import java.util.HashMap;
import java.util.List;
import java.util.Map;

public class App
{
    public static void main(String[] args )
    {
        List<Integer> lista = List.of(1, 2, 3);
        List<Integer> listb = List.of(2, 4);
        Map<String, List<Integer>> map = new HashMap<>();
        map.put("a", lista);
        map.put("b", listb);
        List<Map<String, List<Integer>>> liste = List.of(map);
        // add inline breakpoint here
        liste.stream().map(ele -> ele).forEach(System.out::println);
    }
}

```

For the lambda expression, its signature is `"(Ljava/util/Map<Ljava/lang/String;Ljava/util/List<Ljava/lang/Integer;>;>;)Ljava/util/Map<Ljava/lang/String;Ljava/util/List<Ljava/lang/Integer;>;>;"`, the helper method `toNoneGeneric` converts it to `(Ljava/util/Map;>;)Ljava/util/Map;>;`, which is not correct.

`Signature.getTypeErasure` can handle it well, where it returns `(Ljava/util/Map;)Ljava/util/Map;`.

gayanper: It has dependency to CharOperations. Any reason we keep this later strictly free from JDT. I don't see any usage without JDT and JDT.LS for this extension. I'm I missing something ?

testforstephen: `java.debug.core` is the layer that implements the DAP using Java JDI interface, we don't want it to be mixed with JDT code. I know there are people who reuse this layer for their own Java debugger.

`java.debug.plugin` is the component that integrates it with the jdt.ls.

gayanper: I do understand the abstraction, but I didn't thought that this will be reused for other kind of implementations. Then i think we cannot bring in jdtcore. So can you check the latest. I think its better to duplicate the logics here then, I don't think using the same charops here like in jdt is really needed

testforstephen: There are still cases not covered well in the latest code.

```java
import java.util.HashMap;
import java.util.List;
import java.util.Map;

public class Main {
    public static void main(String[] args )
    {
        List<Integer> lista = List.of(1, 2, 3);
        List<Integer> listb = List.of(2, 4);
        Map<List<Integer>, List<Integer>> map = new HashMap<>();
        map.put(lista, lista);
        map.put(listb, listb);
        List<Map<List<Integer>, List<Integer>>> listc = List.of(map);
        // add inline breakpoint here
        listc.stream().map(ele -> ele).forEach(System.out::println);
    }
}
```

The generic signature is `(Ljava/util/Map<Ljava/util/List<Ljava/lang/Integer;>;Ljava/util/List<Ljava/lang/Integer;>;>;)Ljava/util/Map<Ljava/util/List<Ljava/lang/Integer;>;Ljava/util/List<Ljava/lang/Integer;>;>;`, which is converted to `(Ljava/util/Map;Ljava/util/List;>;)Ljava/util/Map;Ljava/util/List;>;` by `toNoneGeneric`.

A modified version from `Signature.getTypeErasure` can be
```java
    private String toNoneGeneric(String genericSig) {
        int end = genericSig.indexOf('<');
        if (end == -1) return genericSig;
        char[] signature = genericSig.toCharArray();
        int length = signature.length;
        char[] result = new char[length];
        int pos = 0;
        int start = 0;
        int deep= 0;
        for (int idx=end; idx<length; idx++) {
            switch (signature[idx]) {
                case '<':
                    if (deep == 0) {
                        int size = idx-start;
                        System.arraycopy(signature, start, result, pos, size);
                        end = idx;
                        pos += size;
                    }
                    deep++;
                    break;
                case '>':
                    deep--;
                    if (deep < 0) throw new IllegalArgumentException(String.valueOf(signature));
                    if (deep == 0) start = idx+1;
                    break;
            }
        }

        if (deep > 0) throw new IllegalArgumentException(String.valueOf(signature));
        int size = pos+length-start;
        char[] resized = new char[size];
        System.arraycopy(result, 0, resized, 0, pos);
        System.arraycopy(signature, start, resized, pos, length-start);
        return new String(resized);
    }
```

testforstephen: Since the upstream `Signature.getTypeErasure()` is already well-tested, why not use that implementation directly instead of reinventing the wheels?

gayanper: We could just copy it as you suggested, but still we have to change it to avoid dependencies to classes like charoperations. So in stead of copy and change. I thought why not improve what we have here since this variant seems to be much simpler. And I don't see if we should throw errors in this implementation since it's highly unlikely we get a unbalanced signature compared to scenarios upstream methods is used.

testforstephen (approved):
LGTM. thanks for contribution.

PR #500, opened 2023-07-01
Improve inline breakpoint discovery for multi line lambda #489

gayanper:
This pull request improve the discovery of inline breakpoints on lines as described in issue #489. The changes tries to keep the lambda breakpoint location computation as it is while providing the inline breakpoint location based on the formatted code.


gayanper:
@testforstephen WDYT about this change, I closed the previous PR since it could cause side effects

gayanper:
Closing this in favor of https://github.com/microsoft/java-debug/pull/522

PR #501, opened 2023-07-24
Custom request 'refreshFrames' to refresh the thread's frames

testforstephen:


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/RefreshFramesHandler.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public List<Command> getTargetCommands() {
+        return Arrays.asList(Command.REFRESHFRAMES);
+    }
+
+    @Override
+    public CompletableFuture<Response> handle(Command command, Arguments arguments, Response response,
+            IDebugAdapterContext context) {
+        RefreshFramesArguments refreshArgs = (RefreshFramesArguments) arguments;
+        String[] affectedRootPaths = refreshArgs == null ? null : refreshArgs.affectedRootPaths;
```
jdneo, line 42: We can assign affectedRootPaths to an empty string to avoid further null checks

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/RefreshFramesHandler.java
```diff
@@ -0,0 +1,129 @@
...
+            for (String classUri : decompiledClasses) {
+                // decompiled class uri is like 'jdt://contents/rt.jar/java.io/PrintStream.class?=1.helloworld/%5C/usr%5C/lib%5C/jvm%5C/
+                // java-8-oracle%5C/jre%5C/lib%5C/rt.jar%3Cjava.io(PrintStream.class'.
+                if (classUri.startsWith("jdt://contents/")) {
+                    String jarName = classUri.substring("jdt://contents/".length());
+                    int sep = jarName.indexOf("/");
+                    jarName = sep >= 0 ? jarName.substring(0, sep) : jarName;
+                    for (String affectedRootPath : affectedRootPaths) {
+                        if (affectedRootPath.endsWith("/" + jarName) || affectedRootPath.endsWith("\\" + jarName)) {
+                            shouldRefresh = true;
```
jdneo, line 65: Can we directly call `refreshFrames()` and return here?

Or considering extracting the entire for-loop (start from line 56) to another method and early return when affected path is hit.

jdneo (approved):

PR #509, opened 2023-09-07
Add extensibility points to DebugAdapter and ProtocolServer

VladimirMakaev:
## Context

I'm building a Java DAP server based on `com.microsoft.java.debug.core` for our internal projects based on Buck. The goals is to enable debugging without relying on Eclipse LSP. So I'm customizing the part that is resolving source locations based on our internal language tooling and the dependency graph knowledge available in Buck. 

`com.microsoft.java.debug.core` works really well for me as I'm able to completely reuse JDWP <-> DAP interaction while replacing providers with my own implementation. This works quite well.

Additionally our binaries are quite customised so I'd like to also provide custom implementations for "initialize", "launch" and "setFunctionalBreakpoints" requests without necessarily rewriting `ProtocolServer` and `DebugAdapter`

This PR is enabling me to inherit `DebugAdapter` (and inject to `ProtocolServer`) to customize handlers that get executed  without doing weird hacks with reflection or rewriting both of those types

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/DebugAdapter.java
@@ -102,7 +102,7 @@ public class DebugAdapter implements IDebugAdapter {
         }
     }
 
-    private void initialize() {
+    protected void initialize() {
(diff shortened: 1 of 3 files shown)


VladimirMakaev:
@microsoft-github-policy-service agree company="Meta"

testforstephen (approved):

Issue #259, opened 2019-03-18
Could not load source '': SourceRequest: property 'sourceReference' is missing, null, or empty.


cshung:
Cannot load source code during Java debugging in a certain scenario.

##### Environment
- Operating System: OSX
- JDK version: 1.8.0_201
- Visual Studio Code version: 1.30.2
- Java extension version:0.6.0
- Java Debugger extension version:0.17.0

##### Steps To Reproduce
1. Clone and build the google closure compiler at https://github.com/google/closure-compiler
2. Set a breakpoint at src/com/google/javascript/jscomp/PeepholeCollectPropertyAssignments.java
3. Build the project using maven (instruction included at the front page of the repo)
4. Debug

##### Current Result
The breakpoint hits, but then loading sour code fails with this error.
Could not load source '': SourceRequest: property 'sourceReference' is missing, null, or empty.

##### Expected Result
Loading source code should work.

##### Additional Information
This is a clone of [issue 550](https://github.com/Microsoft/vscode-java-debug/issues/550) in the [vscode-java-debug](https://github.com/Microsoft/vscode-java-debug) repo. I believe the root cause of the bug is in this repo, so I cloned the issue over.

I believe I have some idea why the scenario is failing:
[...]

PR #515, opened 2023-10-26
Extend ISourceLookUpProvider to be able to specify whether the source is Local (VSCode) or Remote(DAP Server) 

VladimirMakaev:
## Context

I'm building a DAP server based on `core` package. I.e. without eclipse plugin. Currently it's impossible to set "sourceReference" when returning types in "stacktrace" request. 

Having a distinction between Local and Remote sources allows an implementation of ISourceLookUpProvider to instruct the underlying request handlers whether a follow up with "source" request and have it served via `getSourceContents`

This is also ABI compatible with whoever could implement ISourceLookUpProvider already as current logic is identical to SourceType.Local

## Test Plan
I've tested this works with my DAP server


VladimirMakaev:
This should address #259 without breaking changes in current extension, allowing to have a custom implementation of ISourceLookUpProvider as a solution

testforstephen (approved):

PR #520, opened 2023-11-25
Fix the step-into target on multi line expression. Fix #519

### LINKED ISSUE #519: StepInto method target doesn't work if the method is part of wrapped expression
Take the example

```
mport java.util.ArrayList;
import java.util.Collection;
import java.util.List;

public class DemoApplication {

    new DemoApplication().test();
  }

  private void test() {
    new ArrayList<>(List.of("1"))
      .subList(0, 1)
      .addAll(mergedData());

    new ArrayList<>(List.of("1"))
      .subList(0, 1)
      .addAll(filterMe("1", 
        "2", "3", 
        "4"));

  }

  private Collection<? extends String> mergedData() {
    return List.of("Merged");
  }

  private Collection<? extends String> filterMe(String string1, String string2, String string3, String string4) {
    return mergedData();
  }
}
```

Now try to add breakpoints at expression `new ArrayList<>(List.of("1")).subList(0, 1).addAll(mergedData());` and select `mergedData` as target which will not work.

Also the same if you try on the next expression and try to jump into `filterMe("1", "2", "3", "4")`.

gayanper:
The fix try to compare the current line number against the target MethodInvocation start line if the debugger lines doesn't match because the current frame is at a wrapped line even though we are executing the same expression.

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StepRequestHandler.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StepRequestHandler.java
@@ -451,7 +451,7 @@ public class StepRequestHandler implements IDebugRequestHandler {
         // expression which is wrapped.
         return originalMethod.equals(currentMethod)
                 && (original.lineNumber() == current.lineNumber()
-                        || (targetStepIn != null && targetStepIn.lineStart >= current.lineNumber()));
+                        || (targetStepIn != null && targetStepIn.lineEnd >= current.lineNumber()));
     }
 
     /**
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StepRequestHandler.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StepRequestHandler.java
@@ -268,7 +268,8 @@ public class StepRequestHandler implements IDebugRequestHandler {
                                 } else if (currentStackDepth == threadState.stackDepth) {
                                     // If the ending step location is same as the original location where the step into operation is originated,
                                     // do another step of the same kind.
-                                    if (isSameLocation(currentStepLocation, threadState.stepLocation)) {
+                                    if (isSameLocation(currentStepLocation, threadState.stepLocation,
+                                            threadState.targetStepIn)) {
                                         context.getStepResultManager().removeMethodResult(threadId);
                                         threadState.pendingStepRequest = DebugUtility.createStepIntoRequest(thread,
                                             context.getStepFilters().allowClasses,
@@ -438,15 +439,19 @@ public class StepRequestHandler implements IDebugRequestHandler {
         return true;
     }
 
-    private boolean isSameLocation(Location original, Location current) {
+    private boolean isSameLocation(Location current, Location original, MethodInvocation targetStepIn) {
         if (original == null || current == null) {
             return false;
         }
 
         Method originalMethod = original.method();
         Method currentMethod = current.method();
+        // if the lines doesn't match, check if the current line is still behind the
+        // target if a target exist. This handles where the target is part of a
+        // expression which is wrapped.
         return originalMethod.equals(currentMethod)
-            && original.lineNumber() == current.lineNumber();
+                && (original.lineNumber() == current.lineNumber()
+                        || (targetStepIn != null && targetStepIn.lineStart >= current.lineNumber()));
     }
 
     /**
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/MethodInvocationLocator.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/MethodInvocationLocator.java
@@ -71,7 +71,11 @@ public class MethodInvocationLocator extends ASTVisitor {
 
     @Override
     public boolean visit(TypeDeclaration node) {
-        return shouldVisitNode(node);
+        // when root type declaration doesn't ends with a empty line, the end position
+        // for the root type returned from getLineNumber becomes -1. Therefore the
+        // shouldVisitNode fails, so the extra check unit.getRoot() == node.getRoot() is
+        // added.
+        return unit.getRoot() == node.getRoot() || shouldVisitNode(node);
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


gayanper:
@testforstephen Let me know if you see a better condition to use in this scenario.

testforstephen:
> @testforstephen Let me know if you see a better condition to use in this scenario.

I don't get how this fix works. When I tested the snippet you posted in the issue #519, the `stepInTargets` DAP request returns no results for step-into target operation. Shouldn't we fix MethodInvocationLocator first?

gayanper:
> > @testforstephen Let me know if you see a better condition to use in this scenario.
> 
> I don't get how this fix works. When I tested the snippet you posted in the issue #519, the `stepInTargets` DAP request returns no results for step-into target operation. Shouldn't we fix MethodInvocationLocator first?

Let me check on this, I see the same, probably I missed to commit something

gayanper:
> > > @testforstephen Let me know if you see a better condition to use in this scenario.
> > 
> > 
> > I don't get how this fix works. When I tested the snippet you posted in the issue #519, the `stepInTargets` DAP request returns no results for step-into target operation. Shouldn't we fix MethodInvocationLocator first?
> 
> Let me check on this, I see the same, probably I missed to commit something

Try now @testforstephen

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/MethodInvocationLocator.java
```diff
@@ -71,7 +71,7 @@ public boolean visit(MethodDeclaration node) {
 
     @Override
     public boolean visit(TypeDeclaration node) {
-        return shouldVisitNode(node);
+        return unit.getRoot() == node.getRoot() || shouldVisitNode(node);
```
testforstephen, line 74: The condition check `unit.getRoot() == node.getRoot()` is not needed. I tested #519 snippet without this check, it worked.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StepRequestHandler.java
```diff
@@ -438,15 +439,19 @@ private boolean shouldDoExtraStepInto(int originalStackDepth, Location originalL
         return true;
     }
 
-    private boolean isSameLocation(Location original, Location current) {
+    private boolean isSameLocation(Location original, Location current, MethodInvocation targetStepIn) {
```
testforstephen, line 442: Look at the caller, actually we took the first parameter as `current`, the second one as `original`. So we should change the signature definition to `private boolean isSameLocation(Location current, Location original, MethodInvocation targetStepIn)`.

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StepRequestHandler.java
```diff
@@ -438,15 +439,19 @@ private boolean shouldDoExtraStepInto(int originalStackDepth, Location originalL
...
 
         Method originalMethod = original.method();
         Method currentMethod = current.method();
+        // if the lines doesn't match, check if the current line is still behind the
+        // target if a target exist. This handles where the target is part of a
+        // expression which is wrapped.
         return originalMethod.equals(currentMethod)
-            && original.lineNumber() == current.lineNumber();
+                && (original.lineNumber() == current.lineNumber()
+                        || (targetStepIn != null && targetStepIn.lineStart > current.lineNumber()));
```
testforstephen, line 454: The last condition check should be `(targetStepIn != null && targetStepIn.lineEnd >= current.lineNumber())`, which supports multiline expression better.

testforstephen:
I also found another corner case failure when steping target into List.**of**("1").
```java
new ArrayList<>(List.of("1"))
      .subList(0, 1)
      .addAll(mergedData());
```

It looks like the generic method signature (e.g. "(LE;)Ljava/util/List;") we parsed from JDT is slightly different with the one from the JDI (e.g. "(Ljava/lang/Object;)Ljava/util/List;"). We could fix it in another PR.
https://github.com/microsoft/java-debug/blob/9bdd9977860e5912a39b8efbaf62b9ee5577ec98/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StepRequestHandler.java#L330-L345

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/MethodInvocationLocator.java
gayanper: For some reason for me it start to fail before in `shouldVisitNode` the `end` becomes -1 since the on JDT side the start + length of the node becomes equal to the end of the compilation unit's last position. I will try again without this check.

gayanper: OK now I found the reason, if you don't have a empty line at the end of the class, that after the last class definitions' brace, then the end position becomes -1

testforstephen: I can reproduce it, but i would prefer to modify the implementation `shouldVisitNode()` to fix the problem.

Since the offset range of the ast node is `[node.getStartPosition, node.getStartPosition() + node.getLength() - 1]`, the ending line should be got from `unit.getLineNumber(node.getStartPosition() + node.getLength() - 1)`.

The fix could be as follow:
```java
    private boolean shouldVisitNode(ASTNode node) {
        int start = unit.getLineNumber(node.getStartPosition());
        int end = unit.getLineNumber(node.getStartPosition() + node.getLength() - 1);

        if (line >= start && line <= end) {
            return true;
        }

        return false;
    }
```

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/StepRequestHandler.java
```diff
@@ -438,15 +439,19 @@ private boolean shouldDoExtraStepInto(int originalStackDepth, Location originalL
...
 
         Method originalMethod = original.method();
         Method currentMethod = current.method();
+        // if the lines doesn't match, check if the current line is still behind the
+        // target if a target exist. This handles where the target is part of a
+        // expression which is wrapped.
         return originalMethod.equals(currentMethod)
-            && original.lineNumber() == current.lineNumber();
+                && (original.lineNumber() == current.lineNumber()
+                        || (targetStepIn != null && targetStepIn.lineStart >= current.lineNumber()));
```
testforstephen, line 454: this doesn't work for multiline method invocation. We should use `targetStepIn.lineEnd` in the last check, e.g. `targetStepIn.lineEnd >= current.lineNumber()`.

For the use case below, step-into target `filterMe` doesn't work, it stops at the line of `"2", "3", `.
```java
new ArrayList<>(List.of("1"))
        .subList(0, 1)
        .addAll(filterMe("1", 
            "2", "3", 
            "4"));
```

testforstephen (approved):
I have updated the PR to fix the comments. Now it's good to merge.

PR #522, opened 2023-12-02
Improve inline breakpoint discovery when expression is multiline. Fix #521

### LINKED ISSUE #521: [Critical] Cannot  debug multiline nested method expression's lambda statements
```
package com.example.micrometerdemo;

import com.fasterxml.jackson.annotation.JsonProperty;
import java.lang.reflect.Constructor;
import java.util.Arrays;

public class Snippet {
    private String name;
    

    public Snippet(@JsonProperty("name") String name) {
      this.name = name;
    }


    public final String getName() {
    Constructor<?> constructor =
        Arrays.stream(this.getClass().getDeclaredConstructors())
            .filter(
                input ->
                    input.getParameterAnnotations().length > 0
                        && input.getParameterAnnotations()[0].length > 0
                        && input.getParameterAnnotations()[0][0] instanceof JsonProperty)
            .findFirst()
            .orElseThrow(
                () ->
                    new IllegalStateException(
                        "Constructor must have a first parameter annotated with JsonProperty(\"<operator name>\")"));

    JsonProperty jsonPropertyAnnotation =
        (JsonProperty) constructor.getParameterAnnotations()[0][0];
    return jsonPropertyAnnotation.value();
  }

    new Snippet("am").getName();
  }
}
```

Try to debug the 

```
                  input.getParameterAnnotations().length > 0
                        && input.getParameterAnnotations()[0].length > 0
                        && input.getParameterAnnotations()[0][0] instanceof JsonProperty)
```

By adding a breakpoint at the line `input.getParameterAnnotations().length > 0`, when debugging is started the breakpoint is rejected. But if we try to step into filter method and then step into Predicate#test method the debugger comes into this code block and everything like evaluations works.

This works as expected by adding the line breakpoint as mentioned above in Eclipse.

gayanper:
When checking for lambda expression visible in same line, instead of matching the lambda line to current source line, this fix checks if the lambda is related to the current source line as part of a expression by traversing the parents unit we find the source line.

- Fixes microsoft/vscode-java-debug#1410
- Fixes #521


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
```diff
@@ -325,7 +325,8 @@ private Location findMethodLocaiton(ReferenceType refType, String methodName, St
         for (Method method : methods) {
             if (!method.isAbstract() && !method.isNative()
                     && methodName.equals(method.name())
-                    && (methodSiguature.equals(method.genericSignature()) || methodSiguature.equals(method.signature()))) {
+                    && (methodSiguature.equals(method.genericSignature()) || methodSiguature.equals(method.signature())
+                            || toNoneGeneric(methodSiguature).equals(method.signature()))) {
```
testforstephen, line 329: This is a good catch, it can fix https://github.com/microsoft/vscode-java-debug/issues/1414 as well.

`toNoneGeneric()` function is duplicated with `org.eclipse.jdt.core.Signature.getTypeErasure()`, we can reuse the one from the JDT core directly.

com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -258,6 +258,21 @@ public boolean visit(LambdaExpression node) {
...
+                ASTNode node = lambda;
+                while (node != null) {
+                    int line = astUnit.getLineNumber(node.getStartPosition());
+                    if(line == sourceLine) {
+                        return line;
+                    } else if (line < sourceLine) {
+                        // the lambda doesn't belong to current line at all
+                        break;
+                    }
+                    node = node.getParent();
```
testforstephen, line 272: Recursively checking the parent is not efficient, any way to return earlier?

gayanper: Well do you see that we could run forever where the`line < sourceLine` will be never reached ? Otherwise i think in most cases we will end from the first iteration where `line == sourceLine` right ? Only in a multiline scenario we will traverse to see if the lambda we found belongs to the same line we add a breakpoint. And I also assume this logic will be executed only for breakpoints that are installed on the source line isn't it ?

gayanper: Updated the fix

testforstephen (approved):
LGTM, thanks for contribution.

PR #525, opened 2023-12-08
Update target platform dependencies

mfussenegger:
Fixes build errors. The old location is no longer available:

https://download.eclipse.org/releases/2023-12/202311171000/ returns 404


mfussenegger:
Btw. is there any chance to make sure tagged releases don't depend on any other ephemeral snapshot versions?
E.g. 0.50.0 also doesn't build for me - it seems some resource it depends on also disappeared.

testforstephen:
replaced by https://github.com/microsoft/java-debug/pull/528

testforstephen:
> Btw. is there any chance to make sure tagged releases don't depend on any other ephemeral snapshot versions?
E.g. 0.50.0 also doesn't build for me - it seems some resource it depends on also disappeared.

This is the same challenge for eclipse.jdt.ls, which frequently updates the target platform to get new feature. As a result, the side extension like the java-debug has to update the TP as well to unblock the build job.

PR #529, opened 2023-12-11
Be able to search lambda breakpoint for both chained and multiline lambdas

### LINKED ISSUE #521: [Critical] Cannot  debug multiline nested method expression's lambda statements

testforstephen:
- Fixes microsoft/vscode-java-debug#1410
- Fixes #521

### DIFF
--- a/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
+++ b/com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
@@ -211,7 +211,7 @@ public class JdtSourceLookUpProvider implements ISourceLookUpProvider {
                 // In future, we could consider supporting to update the breakpoint to a valid
                 // location.
                 BreakpointLocationLocator locator = new BreakpointLocationLocator(astUnit,
-                        sourceLine, true, true);
+                        sourceLine, true, true, astUnit.getPosition(sourceLine, 0), 0);
                 astUnit.accept(locator);
                 // When the final valid line location is same as the original line, that
                 // represents it's a valid breakpoint.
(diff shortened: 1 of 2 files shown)


com.microsoft.java.debug.plugin/src/main/java/com/microsoft/java/debug/plugin/internal/JdtSourceLookUpProvider.java
```diff
@@ -211,7 +211,7 @@ public JavaBreakpointLocation[] getBreakpointLocations(String sourceUri, SourceB
                 // In future, we could consider supporting to update the breakpoint to a valid
                 // location.
                 BreakpointLocationLocator locator = new BreakpointLocationLocator(astUnit,
-                        sourceLine, true, true);
+                        sourceLine, true, true, astUnit.getPosition(sourceLine, 0), 0);
```
testforstephen, line 214: This is a limitation of the upstream ValidBreakpointLocationLocator. It requires the selection offset to search for line breakpoints on nested chained or multiline lambdas.

testforstephen:
superseded by #522

Issue #475, opened 2023-03-05
Debugee exit code not correctly reported during event_exited


asmodeus812:
I am having an issue where the exit code of the debugee is reported as 0 instead of 1 when event_exited completes (fires back), when the debugee exits the session. I am using the latest version of this extension and I am using the following dap plugin in neovim https://github.com/mfussenegger/nvim-dap. I did report the issue first there but the maintiner considers this to be an issue within the actual implementation. 

##### Environment
- Operating System: WSL
- JDK version: 19

##### Steps To Reproduce
Initially reported here - https://github.com/mfussenegger/nvim-dap/issues/886
[sample.zip](https://github.com/microsoft/java-debug/files/10891379/sample.zip)

```md
 .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::                (v2.7.1)

2023-03-05 08:59:45.792  INFO 5600 --- [           main] com.shit.TestApplication                 : Starting TestApplication using Java 19.0.1 on Asmodeus with PID 5600 (/h
ome/asmodeus/.projects/terere/target/classes started by asmodeus in /home/asmodeus/.projects/terere)
2023-03-05 08:59:45.794  INFO 5600 --- [           main] com.shit.TestApplication                 : No active profile set, falling back to 1 default profile: "default"
2023-03-05 08:59:46.418  INFO 5600 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Multiple Spring Data modules found, entering strict repository configura
tion mode
2023-03-05 08:59:46.420  INFO 5600 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Bootstrapping Spring Data JPA repositories in DEFAULT mode.
2023-03-05 08:59:46.430  INFO 5600 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Finished Spring Data repository scanning in 4 ms. Found 0 JPA repository
 interfaces.
2023-03-05 08:59:46.468  INFO 5600 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Multiple Spring Data modules found, entering strict repository configura
tion mode
[...]

PR #553, opened 2024-05-19
Propagate the exit code of debugee to exit event

kezhenxu94:
Fixes https://github.com/microsoft/java-debug/issues/475


kezhenxu94:
@testforstephen can you please take a look at this?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/ConfigurationDoneRequestHandler.java
```diff
@@ -82,7 +82,7 @@ private void handleDebugEvent(DebugEvent debugEvent, IDebugSession debugSession,
         } else if (event instanceof VMDeathEvent) {
             vmHandler.disconnectVirtualMachine(event.virtualMachine());
             context.setVmTerminated();
-            context.getProtocolServer().sendEvent(new Events.ExitedEvent(0));
+            context.getProtocolServer().sendEvent(new Events.ExitedEvent(event.virtualMachine().process().exitValue()));
```
testforstephen, line 85: `event.virtualMachine().process()` might return `null`, so a null check is necessary. Additionally, I'm investigating how the exit mechanism works, but it seems my debug session never receives a VMDeathEvent when the process exits. Instead, only a VMDisconnectEvent occurs.

PR #565, opened 2024-08-05
Update lsp4j dependency to version 0.23.1

bsolar17:
New version of jdtls core requires lsp4j 0.23 or greater.

With 0.22.0 the build currently fails with the following error:

```
[ERROR] Cannot resolve target definition:
[ERROR]   Software being installed: org.eclipse.jdt.ls.core 1.39.0.202408021530
[ERROR]   Missing requirement: org.eclipse.jdt.ls.core 1.39.0.202408021530 requires 'osgi.bundle; org.eclipse.lsp4j [0.23.0.0,0.24.0)' but it could not be found
[ERROR]
```


testforstephen (approved):
LGTM. thanks for contribution.

PR #585, opened 2025-04-16
Update target platform to more closely follow JDT-LS.

rgrunber:
- Update Apache Commons IO from 2.11.0 to 2.19.0

java-debug is a bundle contributed into the JDT-LS runtime. This means that it has no control over any dependencies it dos not contribute into the runtime itself (all of them). So when java-debug defines a target platform with dependencies that don't reference the same ones JDT-LS does, it is masking any potential runtime problems. This is the same issue we ran into on LSP4MP / quarkus-ls. The project will compile just fine and fail at runtime because the target platform is not reflective of runtime.

The solution is to reference the exact bundles JDT-LS is using (ie. the latest ones), at https://download.eclipse.org/jdtls/snapshots/repository/latest/plugins/ .

- [ ] Need to verify this by running it but it compiles and tests seem to run just fine.

### DIFF
--- a/com.microsoft.java.debug.plugin/.classpath
+++ b/com.microsoft.java.debug.plugin/.classpath
@@ -7,7 +7,7 @@
 		</attributes>
 	</classpathentry>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry exported="true" kind="lib" path="lib/commons-io-2.11.0.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/commons-io-2.19.0.jar"/>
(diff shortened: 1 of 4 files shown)


svor:
Hi! 👋
@rgrunber  thanks for the upgrading Apache Commons IO library, the old version contains high severity CVE problem. 

@testforstephen  Since this patch is important for the OpenShift Dev Spaces product and directly affects our delivery timeline, could I please ask for a new release of **java-debug** and **vscode-java-debug** with this change included? It would really help us move forward and integrate the fix promptly.

testforstephen:
@svor sure, I will make some test tomorrow. if everything is fine and I will trigger a release for this.

PR #610, opened 2025-10-21
Add suspend mode for source breakpoints

### LINKED ISSUE #608: Support "Suspend VM" and "Suspend Thread" Breakpoint Modes
I propose we support the Java breakpoint suspend modes by adding the "breakpointModes" capability and honoring when breakpoints are set.

tsmaeder:
Fixes #608 

Note that this PR contains code for handling source breakpoints and exception breakpoints. But in fact, VS Code Java only properly handles source and instruction breakpoints. I've left the code in there, since it's not completely trivial and it should not affect the behavior if no suspend mode is sent when setting the exception breakpoints.

**How to Test**
1. Set a breakpoint in a Java Program
2. Start Debugging
3. You now should be able to chose the suspend mode (thread or VM) in the breakpoints view and when editing the breakpoint condition. 
4. Make sure the breakpoint can suspend either all threads or only one thread depending on the mode
5. Put your breakpoint in a for loop and when the breakpoint is hit, switch to the other suspend mode
6. Resume the program (may need "resume all" from the context menu) and ensure the new suspend mode is honored the next time you hit the breakpoint.
7. Make sure exception breakpoints behave as usual. There should be no UI to set the mode and exceptions should only suspend the thread the occur on.

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/Breakpoint.java
@@ -41,6 +41,7 @@ public class Breakpoint implements IBreakpoint {
     private int hitCount = 0;
     private String condition = null;
     private String logMessage = null;
+    private int suspendPolicy = BreakpointRequest.SUSPEND_EVENT_THREAD;
     private HashMap<Object, Object> propertyMap = new HashMap<>();
 
     private boolean async = false;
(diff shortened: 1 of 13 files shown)


tsmaeder:
@chagong is there anything (beyond a rebase) that I could do to move this forward?

tsmaeder:
> @tsmaeder we make it a global flag because vscode doesn't have UI option to set breakpoint property. Are you doing this for cursor?

@chagon VS Code _does_ have UI for this, but you have to enable the modes in the debug adapter (what this PR does) in order to see it. See description.

PR #619, opened 2026-01-21
feat: support suspend all setting

### LINKED ISSUE #722: Pause All Threads on Hitting Breakpoint?
Is it possible to configure a breakpoint (or the debugger in general) to pause **all** threads when a breakpoint is hit?

I have a scenario where once one of my (worker) threads hits a breakpoint, the main thread continues creating more threads that also end up hitting the same breakpoint.  I can manually pause my main thread when my breakpoint is reached, but it would be nice to be able to configure the breakpoint so that _all_ threads are _automatically_ paused.  (This would probably need to be integrated with any new Continue All action being discussed elsewhere to make it easier to resume operation).

chagong:
Fix https://github.com/microsoft/vscode-java-debug/issues/722

### DIFF
--- a/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/DebugSession.java
+++ b/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/DebugSession.java
@@ -264,7 +264,7 @@ public class DebugSession implements IDebugSession {
     }
 
     @Override
-    public boolean suspendAllThreads() {
+    public boolean shouldSuspendAllThreads() {
         return suspendAllThreads;
     }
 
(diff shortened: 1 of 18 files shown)


com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/adapter/handler/ThreadsRequestHandler.java
```diff
@@ -149,6 +149,11 @@ private CompletableFuture<Response> resume(Requests.ContinueArguments arguments,
         if (thread == null) {
             thread = DebugUtility.getThread(context.getDebugSession(), arguments.threadId);
         }
+
+        if (context.getDebugSession().suspendAllThreads()) {
```
testforstephen: what's the purpose of this code?

com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/DebugSession.java
```diff
@@ -260,17 +263,22 @@ public VirtualMachine getVM() {
         return vm;
     }
 
+    @Override
+    public boolean suspendAllThreads() {
```
testforstephen: Could we rename this to `shouldSuspendAllThreads()`? The current name sounds like an operation instead of a flag, especially since the class already has operations like suspend and getAllThreads.

testforstephen (approved):

aikee07 (approved):

aikee07 (approved):

